PR #1, opened 2017-07-11
Add sg.conf to force load 'sg' module at boot time

fbuihuu:
It's currently shipped by systemd but it's more appropriate to ship
this via a package maintained by the kernel team, see bsc#1036463.


mwilck:
Superseded this with commit 72d3456.

PR #5, opened 2019-01-18
Add file system blacklist file (fate#326832).

jeffmahoney:
As automated fuzzer testing grows in popularity, we've seen an uptick of security vulnerabilities that are in rarely used file systems. Normally this isn't an issue but the combination of user-mountable file systems and the automatic mounting of file systems on removable devices means that unprivileged users can cause the automatic loading of kernel modules to handle those operations. This feature blacklists uncommon file systems by default and allows the user to enable them if their use requires them.

### DIFF
--- a/modprobe.conf/modprobe.conf.fs-blacklist
+++ b/modprobe.conf/modprobe.conf.fs-blacklist
@@ -1,0 +1,32 @@
+# These file systems are not used on most systems but the kernel
+# will still happily load the modules for them when presented with
+# a potentially malicious device like a USB stick.  Many of these
+# have not been maintained except for in-kernel API compatibility
+# in many years, have not been audited, and may have security vulnerabilities.
+#
+# The following list only specifies local file systems since those are the
+# only ones that will be detected automatically by mount(8).
+#
+# Enable at your own risk.
+#
+blacklist adfs
+blacklist affs
+blacklist bfs
+blacklist befs
+blacklist cramfs
+blacklist efs
+blacklist erofs
+blacklist exofs
+blacklist freevxfs
+blacklist f2fs
+blacklist hfs
+blacklist hpfs
+blacklist jffs2
+blacklist jfs
+blacklist minix
+blacklist nilfs2
+blacklist qnx4
+blacklist qnx6
+blacklist sysv
+blacklist ubifs
+blacklist ufs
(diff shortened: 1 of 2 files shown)


mwilck:
@jeffmahoney, I believe this makes only sense for SLES, correct?

I'd expect an uproar for openSUSE, as the list contains some filesystems that might be relevant to home users (cramfs is even used by e.g. the SUSE mkdud tool, hfs may be necessary for Mac users (?))

It's been requested for SLE15-SP1, so I'll enable this for SLE only for the time being.

solardiz:
I'd recommend a whitelist of trusted filesystems, but I guess that'd be much harder to implement?

modprobe.conf/modprobe.conf.fs-blacklist
```diff
@@ -0,0 +1,32 @@
...
+blacklist befs
+blacklist cramfs
+blacklist efs
+blacklist erofs
+blacklist exofs
+blacklist freevxfs
+blacklist f2fs
+blacklist hfs
+blacklist hpfs
+blacklist jffs2
```
richardweinberger: This is a raw flash filesystem, it does not work on a block device. So there is no way to have this automatically mounted by plugging in a USB thumb drive.

modprobe.conf/modprobe.conf.fs-blacklist
```diff
@@ -0,0 +1,32 @@
...
+blacklist jfs
+blacklist minix
+blacklist nilfs2
+blacklist qnx4
+blacklist qnx6
+blacklist sysv
+blacklist ubifs
```
richardweinberger: Same as for jffs2. But UBIFS is actively developed and maintained.
I can understand that OpenSUSE blacklists flash filesystems because you need them only on embedded
devices with raw flash. But please don't claim that they are unmaintained/vulnerable.

jeffmahoney: The same applies to ubifs.  But your more general point is well taken. You're right. Martin Wilck has a newer version of this that creates one file per blacklist entry and has a comment at the top of each.  Currently it uses the phrasing " because it isn't actively supported by SUSE, not well maintained, and may have security vulnerabilites."

It would make more sense to change the phrasing to "or" instead of "and."

modprobe.conf/modprobe.conf.fs-blacklist
jeffmahoney: True, and the code doesn't allow loopback mounts.  We can drop jffs2.

PR #9, opened 2019-05-31
weak-modules2: Fix spurious depmod error when removing KMP (bsc#1127155).

hramrach:
The KMP to be removed can be replaced with another KMP later.

Running depmod before a replacement is found can lead to spurious error
mesasges.


hramrach:
Fixed by 61926891354a2a1ca6cd73a098458723165677da

PR #10, opened 2019-11-06
modprobe.conf: add dependency of papr_scm on libnvdimm

hramrach:
(bsc#1142152, ltc#176292, FATE#327775).

libnvdimm is not included in the initramfs automatically.


mwilck:
The FATE is for SLE15-SP1. Is it sufficient to add this to SLE15-SP1 and later?

hramrach:
And SLE12 SP5 as well.

Thanks

PR #12, opened 2020-02-27
Replace use of rpmsort with custom luarpmsort

Vogtinator:
Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1164553

Warning: COMPLETELY UNTESTED! Please tell me how to test this and I'll try to do that.

An alternative to using this new `luarpmsort` script would be to just ship a copy of the old `/usr/lib/rpm/rpmsort`, but that doesn't handle `^` and `~` properly.

### DIFF
--- a/driver-check.sh
+++ b/driver-check.sh
@@ -301,7 +301,7 @@ check_rpm "$smt"
 mkdir -p "$tmp/rpms"
 found_kernel=false
 for rpm in $(rpm -qa --qf '%{n}-%{v}-%{r}\n' 'kernel-*' '*-kmp-*' | \
-		/usr/lib/rpm/rpmsort); do
+		"$(dirname "$0")/luarpmsort" | tac); do
 	case "$rpm" in
 	kernel-source-* | kernel-syms-* | kernel-*-debug* | kernel-*-man-* | \
 	kernel-*-devel-* | kernel-firmware-* | kernel-coverage-* | \
--- a/luarpmsort
+++ b/luarpmsort
@@ -1,0 +1,44 @@
+#!/bin/bash
+rpmsort() {
+	script='
+function parse(ver)
+	local epoch, version, release = 0, ver, 0
+	_, eend, e = ver:find("^(%d+):")
+	if eend then
+		ver = ver:sub(eend + 1)
+		version = ver
+		epoch = e
+	end
+	_, _, v, r = ver:find("(.+)%-(.+)$")
+	if v then
+		version = v
+		release = r
+	end
+	return epoch, version, release
+end
+
+function pkgvercmp(a, b)
+	local ae, av, ar = parse(a)
+	local be, bv, br = parse(b)
+
+	local ecmp = rpm.vercmp(ae, be)
+	if ecmp ~= 0 then return ecmp end
+
+	local vcmp = rpm.vercmp(av, bv)
+	if vcmp ~= 0 then return vcmp end
+
+	return rpm.vercmp(ar, br)
+end	
+
+vers = {}
+for line in io.stdin:lines() do
+	table.insert(vers, line)
+end
+table.sort(vers, function(a, b) return pkgvercmp(a, b) == -1 end)
+print(table.concat(vers, "\n"))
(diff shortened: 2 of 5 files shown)


mwilck:
AFAICS, your code is fine. I've given it some basic testing.

That said, rpmsort is abused in suse-module-tools. In weak-modules2, it's passed a list of installed KMP packages, like this:

    vhba-kmp-default-20190410_k5.2.8_1-1.5
    vhba-kmp-default-20190410_k5.2.1_1-1.4
    qla2xxx-kmp-default-10.01.00.20_k5.2.8_1-1.10
    vhba-kmp-default-20190410_k5.4.14_1-1.27

(the "flavor", here "default" is always the same).

Yes, the package names are included in the input and would thus be interpreted as part of the "epoch" or "version". But I guess this is intended, this way the kmps will get sorted alphabetically by name, and by version/release. The "k" makes sure that vercmp will get this right.

In `weak-modules`, it's even worse.  The second and third call are used to sort the output of `ls /lib/modules`, and all directories there end in `-$flavor`, so both the old and new script will assume that the "release" is `-default`. The same holds for the first call, but there only two versions are compared, so it's basically a `rpm.vercmp()` call only.

But weak-modules has been moved to suse-module-tools-legacy anyway a year ago, and can be removed in factory. IIRC it hasn't been used since SLE10.

So, what I'm going to do is ditch weak-modules, and insert your function directly into weak-modules2. OK?

mwilck:
`weak-modules` has never been shipped  in SLE15 or later.

Vogtinator:
> That said, rpmsort is abused in suse-module-tools. In weak-modules2, it's passed a list of installed KMP packages, like this:

Except for `~` and `^` characters in versions, the behaviour should be the same as `rpmsort` was, so whatever hack was used there should still continue to work fine...

> Yes, the package names are included in the input and would thus be interpreted as part of the "epoch" or "version". But I guess this is intended, this way the kmps will get sorted alphabetically by name, and by version/release. The "k" makes sure that vercmp will get this right.

It would treat the name as part of the version, e.g.
`vhba-kmp-default-20190410_k5.4.14_1-1.27` -> epoch: `0` version: `vhba-kmp-default-20190410_k5.4.14_1` release: `1.27`

> So, what I'm going to do is ditch weak-modules, and insert your function directly into weak-modules2. OK?

Sounds good to me.

mwilck:
> > Yes, the package names are included in the input and would thus be interpreted as part of the "epoch" or "version". [...]
> 
> It would treat the name as part of the version, [...]

What I meant was: *if* there was an epoch, the filename would be taken to be part of the epoch.
But we're not using the epoch for kernels of course, and it doesn't matter anyway.

mwilck:
Notes: 

 1. I have to confess I don't know what effect `^` has on rpm sorting. 
 2. There's also `rpmdev-sort` from the `rpmdevtools` package, but it's written in python, and we don't want suse-module-tools to depend on python. Moreover, it works only if `%{name}-%{version}-%{release}` is passed in.

Vogtinator:
> Given that rpmsort was called with "-r" in all but one place,

Two places actually, `driver-check.sh` uses it as well (even though I'm not sure whether it has any noticeable impact)

> could you change luarpmsort such that it sorts backwards, and remove the "tac" calls?

I changed the comparison so that only `driver-check.sh` uses `tac` now.

mwilck:
> Two places actually, `driver-check.sh` uses it as well (even though I'm not sure whether it has any noticeable impact)

Argh, seems that we both overlooked this the first time around. Thanks for spotting it. Actually, this has no impact whatsoever except for the output ordering. We might as well remove the call there, or replace it by a simple "sort", nobody is going to even notice (this tool was utterly broken for years until I fixed it about a year ago, and nobody complained about it).

mwilck:
@Vogtinator, 

I've modified this yet a bit, by implementing "-r" in luarpmsort. Please check https://github.com/openSUSE/suse-module-tools/pull/13. I don't seem to be able to add you as a reviewer directly.

PR #13, opened 2020-03-05
luarpmsort: add -r option

mwilck:
This allows us to use luarpmsort like the legacy rpmsort program.
Also, set RPMSORT variable once in all scripts that use it.

### DIFF
--- a/luarpmsort
+++ b/luarpmsort
@@ -1,6 +1,6 @@
 #!/bin/bash
 rpmsort() {
-        direction=$1
+	direction=$1
 	script='
 function parse(ver)
 	local epoch, version, release = 0, ver, 0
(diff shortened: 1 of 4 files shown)


luarpmsort
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-# Note: Unlike the old rpmsort, newer versions are at the top!
 rpmsort() {
+        direction=$1
```
Vogtinator: Tabs vs. spaces

luarpmsort
```diff
@@ -29,17 +29,38 @@ function pkgvercmp(a, b)
...
 print(table.concat(vers, "\n"))
 '
 
 	rpm --eval "%{lua: ${script}}"
 }
 
-rpmsort
+usage() {
+    cat >&2 <<EOF
+$0 - sort standard input according to rpm version sorting conventions,
```
Vogtinator: As there's a proper `--help` option now, let's also make it more useful than rpmsort's ever was and add something like `Expects RPM package versions separated by newlines as input and outputs it according to rpm version sorting conventions with old versions at the top.`

mwilck: OK, done. Please check again.

PR #14, opened 2020-05-04
Remove more SUSE INITRD instructions

danimo:
- crc32c is now an implicit softdep of libcrc32
- cifs automatically has the relevant dependencies, as does the dracut module
- libcrc32c now has crc32c as a soft-pre in its own right
- nls is compiled in, but nls_utf8 is still needed. However, it
  can be provided as a pure soft-pre now

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1171065.

### DIFF
--- a/modprobe.conf/modprobe.conf.common
+++ b/modprobe.conf/modprobe.conf.common
@@ -71,18 +71,10 @@ blacklist pata_acpi
 softdep uhci-hcd pre: ehci-hcd
 softdep ohci-hcd pre: ehci-hcd
 
-# libcrc32c calls crypto_alloc_shash("crc32c", 0, 0), which results in a
-# request_module("crc32c"), but that dependency is not seen by modpost/depmod
-# https://bugzilla.novell.com/552443
-# SUSE INITRD: libcrc32c REQUIRES crc32c
-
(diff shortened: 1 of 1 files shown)


modprobe.conf/modprobe.conf.common
```diff
@@ -71,18 +71,10 @@ blacklist pata_acpi
...
-
-# likewise, cifs calls crypto_alloc_shash(...) for a couple of algorithms
-# https://bugzilla.novell.com/730617
-# SUSE INITRD: cifs REQUIRES des ecb md4 md5 hmac arc4
-# It also calls load_nls()
+# cifs calls load_nls()
 # FIXME: determine the proper nls module based on fstab options instead
-# SUSE INITRD: cifs REQUIRES nls_utf8
- 
+softdep nls pre: cifs: nls_utf8
```
mwilck, line 76: This looks wrong. Did you mean `softdep cifs pre: nls_utf8` ?

hramrach: yes, that sounds like it. 
at any rate cifs: is invalid syntax and nls is built-in so the rule would never apply

mwilck: I've reverted this hunk after the merge.

PR #15, opened 2020-07-21
spec: rework dependencies

thkukuk:
Make sure required packages for %post install are already installed at that
time. Use file requires where possible to be able to reduce system size for
small OS.

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -36,12 +36,15 @@ Group:          System/Base
 URL:            https://github.com/openSUSE/suse-module-tools
 Source0:        %{name}-%{version}.tar.xz
 Source1:        %{name}.rpmlintrc
+Requires:       /usr/bin/grep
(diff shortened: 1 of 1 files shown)


mwilck:
Would it help you further if we got rid of "grep" in `%post` altogether?

thkukuk:
@mwilck the best %pre/%post install sections are the one which don't exist ;)
grep, sed, rm, mkdir, cp are commands, which are always there as you always need them. So the usage is Ok. For minimized images like Carwos and maybe MicroOS we only prefer to use busybox and not coreutils and similar packages, as nobody really works on such installations and size matters.

mwilck:
Thanks. I thought about `grep` because dracut doesn't package it by default, replacing it with `sed` calls, and that works quite well once you get used to it. But if you install `grep` anyway, fine. Wrt `%post` scripts in general, it seems that we are carrying around a lot of legacy that nobody dares to remove...

PR #16, opened 2020-09-21
Disable Coldplug for memory devices on ppc64-le, PowerVM

mwilck:
The idea is to exclude memory devices in the standard `systemd-udev-trigger.service` (as we did before, but using a drop-in rather than modifying the unit file), and add another service, tied to `systemd-udev-trigger`, that starts only coldplug for the memory devices but is subject to `ConditionVirtualization` and `ConditionArchitecture` directives.

On PowerVM:

```
jostaberry-6:~ # systemctl show -p ConditionResult -p PartOf systemd-udev-trigger-memory-devices.service
PartOf=systemd-udev-trigger.service
ConditionResult=no
```

On KVM:
```
shiraz-1:~/:[0]# systemctl show -p ConditionResult -p PartOf systemd-udev-trigger-memory-devices.service
PartOf=systemd-udev-trigger.service
ConditionResult=yes
```


hramrach:
how is this supposed to work?

ExecStart=/usr/bin/udevadm trigger --type=subsystems --action=add ; /usr/bin/udevadm trigger --type=devices --action=add --subsystem-nomatch=memory

This looks like it executes the coldplug twice - once with memory devices and once without

hramrach:
Can you change the condition to trigger when KVM virtualization is in effect?

Currently no virtualization can mean either PowerVM or powernv but this is a bug - PowerVM is pHyp virtualization, and once systemd is fixed to detect that the conditiln will break, right?

mwilck:
@hramrach,

> ExecStart=/usr/bin/udevadm trigger --type=subsystems --action=add ; /usr/bin/udevadm trigger --type=devices --action=add --subsystem-nomatch=memory
> 
> This looks like it executes the coldplug twice - once with memory devices and once without

First it triggers `--type=subsystems`. Don't ask me why this is done, but I'm sure it does *not* trigger any device uevents. The second command explicitly skips uevents for memory devices. Note the `ExecStart=` line in the drop-in file - it resets `ExecStart` to an empty list, so that `ExecStart` line from the original `systemd-udev-trigger.service` won't be run. The memory device uevents will be triggered by the additional service *only*. Makes sense?

> Can you change the condition to trigger when KVM virtualization is in effect?

I sure could. Currently the idea is to restrict the set of scenarios where this workaround is applied (i.e. where the additional service is *not* run) to just that one scenario where we deem it necessary (PowerVM).

> Currently no virtualization can mean either PowerVM or powernv but this is a bug - PowerVM is pHyp virtualization, and once systemd is fixed to detect that the conditiln will break, right?

True. But is a fix coming? Once we have a fix in systemd, we can change the condition. Unfortunately, it might then be hard to express the actual logic with systemd's `ConditionVirtualization` directives, as they aren't suitable for complex logic. You can't say "everything except bare metal and PowerVM", for example.

hramrach:
> > Currently no virtualization can mean either PowerVM or powernv but this is a bug - PowerVM is pHyp virtualization, and once systemd is fixed to detect that the conditiln will break, right?
> 
> True. But is a fix coming? Once we have a fix in systemd, we can change the condition. Unfortunately, it might then be hard to express the actual logic with systemd's `ConditionVirtualization` directives, as they aren't suitable for complex logic. You can't say "everything except bare metal and PowerVM", for example.

But saying 'run this only on KVM' should be possible and future-proof.

mwilck:
> But saying 'run this only on KVM' should be possible and future-proof.

I you say it, fine with me ... what about "qemu", for instance?

hramrach:
> > But saying 'run this only on KVM' should be possible and future-proof.
> 
> If you say it, fine with me ... what about "qemu", for instance?

Is that a separate virtualization type?
The system should not be able to tell if it runs under KVM or the CPU is emulated.

Maybe we should fix virt-detect first then?

hramrach:
Actually, running under qemu without KVM does not work whether it can be detected as separate virtualization type or not. I don't think we care about memory hotplug on powernv but the memblock devices might still be there. Detecting KVM should be fine.

mwilck:
> Is that a separate virtualization type?

Yes. [systemd-detect-virt(1)](https://www.freedesktop.org/software/systemd/man/systemd-detect-virt.html), [src/basic/virt.c](https://github.com/systemd/systemd/blob/master/src/basic/virt.c).

> The system should not be able to tell if it runs under KVM or the CPU is emulated.

Apparently it can.

> Maybe we should fix virt-detect first then?

I doubt we can do much about this. It has been like that for a while, and it's documented. It might actually be useful for some use cases. 
The PowerVM detection is a different thing: if we can point out a reasonably clean way to detect it, I guess it could be added.

This is actually why I decided to use `ConditionVirtualization=yes` and not `ConditionVirtualization=kvm` - the list of possible return values of `systemd-detect-virt` is long, and I neither know which ones are available on PPC, nor on which ones memory devices exist, let alone on which ones we'd want to see the coldplug events for them. I thought I was on the safe side with `yes`.

Maybe we shouldn't carry this too far, I believe that for practical purposes,  `ConditionVirtualization=yes`  and `ConditionVirtualization=kvm` are pretty much equivalent (as long as the PowerVM detection isn't fixed).

@fbuihuu, opinions on the latter?

mwilck:
> Detecting KVM should be fine.

Ok, I'll change that, then. Thanks.

hramrach:
The systemd virtualization detection is quite misguided. For one it goes out of its way to detect Xen dom0 as non-virtualized. Unfortunately, we can't rely on anything else being present.

fbuihuu:
This seems a bit overly complex... why don't you simply write a generator which would create a drop-in which in its turn would override `ExecStart=` option of `systemd-udev-trigger.service` to skip memory devices only when it's necessary ?

fbuihuu:
Also I'm not sure to see the point of bumping twice the version of this package... I mean you can add any additional commits to polish your work and then bump the version only once.

mwilck:
@fbuihuu,

thanks for looking at this.

> This seems a bit overly complex... why don't you simply write a generator which would create a drop-in which in its turn would override `ExecStart=` option of `systemd-udev-trigger.service` to skip memory devices only when it's necessary ?

Interesting idea, but that doesn't sound less complex to me in the first place. And more importantly, it would be less transparent to the user, AFAICS.
I guess it's a matter of taste ... I'd need to see a proof of concept to be convinced.

> Also I'm not sure to see the point of bumping twice the version of this package... I mean you can add any additional commits to polish your work and then bump the version only once.

Never mind that. It was because of @hramrach's suggested change, and because I had included the version bump in the PR already (which was stupid, I admit it). I'll consider can rebasing this before it's merged.

fbuihuu:
> Interesting idea, but that doesn't sound less complex to me in the first place.

It is less complex as with the generator you don't need to care about the dependencies.

> And more importantly, it would be less transparent to the user, AFAICS.

What do you mean ?

mwilck:
> > And more importantly, it would be less transparent to the user, AFAICS.
> 
> What do you mean ?

I find generated units (other than mount units and tmpfiles.d, to which we've got used over the years) rather confusing. You see an `ExecStart=` line in the unit file, and you ask yourself why systemd executes something different. IMHO it's mostly a feature for die-hard systemd freaks. Just my personal feeling. Maybe it's just me.

Also, I've never written a generator so far.

fbuihuu:
> I find generated units (other than mount units and tmpfiles.d, to which we've got used over the years) rather confusing. You see an `ExecStart=` line in the unit file, and you ask yourself why systemd executes something different.

`systemctl cat foo.service` should show you all fragments used to make up a unit file.

> IMHO it's mostly a feature for die-hard systemd freaks. Just my personal feeling. Maybe it's just me, and maybe I don't fully understand your suggestion.

Not really, generators have been introduced to overload units according to the runtime environment so it perfectly fits with your use case. So you only need to create a drop-in in case the memory devices should be skipped.

No more headache to get dependencies right, especially when `systemd-udev-trigger.service` is restarted or disabled or whatever.

> Also, I've never written a generator so far.

See man systemd.genertor, it's simple really.

mwilck:
I can see the generator idea has benefits. Not sure if I'll be able to change it in time, given that the current approach "works" as well, and that we should really move forward with this soon.

mwilck:
OK @fbuihuu, please have another look. Using a generator now. It took a few attempts because I didn't realize I couldn't use a here document without changing TMPDIR. Anyway, it seems to work now.

udev-trigger-generator
```diff
@@ -0,0 +1,27 @@
+#! /bin/bash
+# Copyright (c) 2020 SUSE LLC
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# bsc#1174287: On PowerPC, exclude memory devices from udev coldplug
+export PATH=/usr/bin
```
fbuihuu, line 6: I don't think that's needed, PID1 is supposed to start generators with a clean env.

udev-trigger-generator
```diff
@@ -0,0 +1,27 @@
...
+# /tmp may not be writable yet
+export TMPDIR=/dev/shm
```
fbuihuu, line 8: why do you need to export TMPDIR ?

udev-trigger-generator
```diff
@@ -0,0 +1,27 @@
...
+trap 'echo $0: error in "$BASH_COMMAND" >&2' ERR
+
+[[ $(uname -m) = ppc64le ]] || exit 0
+case $(systemd-detect-virt) in
+    none|powervm) ;;
+    *) exit 0;;
+esac
+
+DIR=$1
+[[ $DIR ]]
```
fbuihuu, line 18: what is it for ?

udev-trigger-generator
```diff
@@ -0,0 +1,27 @@
...
+DROPIN="$DIR/systemd-udev-trigger.service.d"
+mkdir -p "$DROPIN"
+cat >"$DROPIN/memory-devices.conf" <<EOF
```
fbuihuu, line 21: perhaps include a prefix which encodes the priority of the drop-in, something like `30-skip-memory-devices.conf` ?

mwilck:
Thanks, @fbuihuu. I pushed another version. In particular, I added a priority for the drop-in, and some comments to explain the code better.

OK like this?

mwilck (commented):
I opened this review by mistake. Closing it.

udev-trigger-generator
fbuihuu: well if that's not needed, there's no point to keep it as it's just confusing otherwise.

udev-trigger-generator
fbuihuu: I don't think it tests for the presence of a directory. `[[ $DIR ]]` only fails if $DIR is the empty string, which can't happen.

mwilck: You're right, sorry. Still warming up after vacation ... I'll change this to 

    [[ $DIR  && -d $DIR ]]

fbuihuu:
@mwilck just in case you missed them, I provided some feedback - basically I don't think you should put extra code if they are not needed as it might confuse the reader.

That was just my 2 cents, feel free to ignore and commit.

udev-trigger-generator
mwilck: I disagree. This way the code can be read without prior knowledge about the environment which PID 1 may or may not have set up. The man page **systemd.generator(7)** doesn't say anything about the environment in which the generators are run.

PR #17, opened 2021-04-07
Enable f2fs

goldwynr:
f2fs was excluded because the on-disk format kept changing and is not
versioned. The on-disk format seems stable for now and can be enabled.


mwilck:
Looking good, thanks. I'm assuming that we still don't support f2fs in SLE, right?

PR #18, opened 2021-04-20
dm-crypt requires essiv in SLE15 SP3 (boo#1183063 bsc#1184134).

hramrach:


mwilck:
Looks good,  thanks. I'll push this for Factory and SLE15-SP3.

PR #19, opened 2021-05-21
sunsetting mkinitrd

mwilck:
Have suse-module-tools work without a `mkinitrd` tool.

@danimo' post to opensuse-factory, 2021-05-18

with the pending submission of Dracut 054, which removes mkinitrd from
the upstream sources, it's time to finally remove mkinitrd in Factory, too:

Changes starting with 054
-------------------------

- mkinitrd will temporarily live in its own dracut-mkinitrd-deprecated
package. It will still be installed when requiring "mkinitrd" and
requires dracut, so nothing changes here.
- If you only required dracut, but meant to require mkinitrd, you need
to change this temporarily or change your scripts to directly call dracut.

Rationale
---------

mkinitrd has always been a rather mediocre wrapper that claimed to keep
compatibility with the SLES 11 implementation. However:

- it has a lot of bugs, quite some documented (!) flags have never
worked, others are marked unimplemented or will say so only if invoked.
I see little point in doing that given
- it adds little value over just using plain Dracut

Differences in calling dracut over calling mkinitrd:
----------------------------------------------------

- Dracut only updates the initrd for the currently running kernel,
unless either a kernel version or --regenerate-all is called
- mkinitrd updates the bootloader by default. Dracut has no such option.
Then again, those are two different concerns that are only merged into a
single command for convenience. Usually this should be invoked from the
kernel install script.
- mkinitrd parses the no longer existing /etc/sysconfig/kernel file.
- mkinitrd handles xen-specific modules. I have not enough understanding
of the state of Xen in Factory to judge whether this is still required.
If so, it should be handled in a dracut module. However, judging by the
fact that the default is sourced from the defunc'ed kernel sysconfig
file, this is probably also dead.
- mkinitrd has slightly different semantics on how to specify a boot
device for NFS & Co.

Point of Contact
----------------

If you have concerns how this affects your package, please let us (me,
Thomas Blume or Neil) know. For the Kernel and other packages that are
definitely impacted once mkinitrd would go away (e.g. YaST), we will
reach out separately and / or file bugzilla issues.

>


hramrach:
we should also drop modutils - see #21

hramrach:
note: there is slight semantic difference between `mkinitrd` and `dracut -f --regenerate-all`. mkinitrd only regenerates existing initrds and does not create initrd for ramdiskless non-modular kernel. This is the reason why I never attempted to do this.

Even if you disregard this mkinitrd does by default the thing that's needed but you must pass two arguments to dracut. Why must all the SUSE tools have so jarring interface?
There is no command for regenerating the bootloader configuration without supplying some arcane arguments either.

mwilck:
> note: there is slight semantic difference between `mkinitrd` and `dracut -f --regenerate-all`. mkinitrd only regenerates existing initrds and does not create initrd for ramdiskless non-modular kernel. This is the reason why I never attempted to do this.

Good to know. I haven't seen a ramdiskless system for ages. Does that do any harm for ramdiskless setups (in the sense that the system wouldn't boot, or similar severe issues)?

> Even if you disregard this mkinitrd does by default the thing that's needed but you must pass two arguments to dracut. Why must all the SUSE tools have so jarring interface?
> There is no command for regenerating the bootloader configuration without supplying some arcane arguments either.

I don't understand what you are saying here. Which tool do cou consider jarring and arcane - dracut or mkinitrd, or both?

hramrach:
On Tue, May 25, 2021 at 01:08:28AM -0700, Martin Wilck wrote:
> > note: there is slight semantic difference between `mkinitrd` and `dracut -f --regenerate-all`. mkinitrd only regenerates existing initrds and does not create initrd for ramdiskless non-modular kernel. This is the reason why I never attempted to do this.
> 
> Good to know. I haven't seen a ramdiskless system for ages. Does that do any harm for ramdiskless setups (in the sense that the system wouldn't boot, or similar severe issues)?
I haven't either, and if mkinitrd was internally dracut they are
probably not supported anyway.
> 
> > Even if you disregard this mkinitrd does by default the thing that's needed but you must pass two arguments to dracut. Why must all the SUSE tools have so jarring interface?
> > There is no command for regenerating the bootloader configuration without supplying some arcane arguments either.
> 
> I don't understand what you are saying here. Which tool do you consider jarring and arcane - dracut or mkinitrd, or both?
With mkinitrd the package postinstall script or the kernel makefile will
call mkinitrd to create the ramdisk for the newly installed kernel, and
when you need to rebuild ramdisks you just call `mkinitrd`. In the case
your installation scripts failed you have to specify the kernel version
once to change from no ramdisk to having ramdisk for your newly
installed kernel.

Compared to that if you call `dracut` under normal circumstances it does
nothing. The default is pretty useless. Every time you have to specify
some arguents for it to be useful.

Similarily there is update-bootloader or similar tool that refreshes
your bootloader configuration, without any arguments it just does it.

On SUSE you have to call `grub2-mkconfig` and redirect the output to the
correct file, or use the undocumented `pbl --config`.

mwilck:
> Compared to that if you call `dracut` under normal circumstances it does nothing. The default is pretty useless.

Ack. But that's really an upstream thing, upstream meaning Fedora in this case. used to think  SUSE's `mkinitrd` was superior to dracut, and wondered why it was ditched, but we can't do anything about that any more.

> Similarily there is update-bootloader or similar tool that refreshes your bootloader configuration, without any arguments it just does it. On SUSE you have to call `grub2-mkconfig` and redirect the output to the correct file, or use the undocumented `pbl --config`.

To be fair, `update-bootloader` (which I assume is from Debian?) is available on SLE.

mwilck:
So @hramrach, can I take your remarks as a positive review, or are you requesting changes?

mwilck:
Btw, in the past you had to call dracut with at least *two* arguments:

    dracut /boot/initrd-$SOME-KERNEL-VERSION $SOME-KERNEL-VERSION

Luckily today dracut is able to guess one from the other if only one is given.

hramrach:
On Tue, May 25, 2021 at 02:23:06AM -0700, Martin Wilck wrote:
> So @hramrach, can I take your remarks as a positive review, or are you requesting changes?

I don't think there is anything to be done about this in
suse-module-tools. The damage has been done by migrating from mkinitrd
to dracut long time ago.

hramrach:
On Tue, May 25, 2021 at 02:21:51AM -0700, Martin Wilck wrote:

> > Similarily there is update-bootloader or similar tool that refreshes your bootloader configuration, without any arguments it just does it. On SUSE you have to call `grub2-mkconfig` and redirect the output to the correct file, or use the undocumented `pbl --config`.
> 
> To be fair, `update-bootloader` (which I assume is from Debian?) is available on SLE.

There is only `update-bootloader-rpm-macros` which contains only 1 file
`/usr/lib/rpm/macros.d/macros.update-bootloader` so it does not sound
like it. Also it would be better to document and promote and add
bash-completion for pbl if that's the tool of choice.

After all `update-bootloader` works great on Debian because there are
many update-XY tools, some upstream and some from Debian. You don't need
bash-completion update-bootloader because it does the right thing by
default, without arguments and any trivial shell can complete command
names.

danimo (approved):
LGTM, thanks!

mwilck:
> There is only `update-bootloader-rpm-macros` which contains only 1 file `/usr/lib/rpm/macros.d/macros.update-bootloader` so it does not sound like it.

I'm on SLE15-SP3 and I have here:

```
bremer:~/:[0]# type update-bootloader
update-bootloader is /sbin/update-bootloader
bremer:~/:[0]# rpm -qf /sbin/update-bootloader
perl-Bootloader-0.933-1.2.x86_64
```

I'm not aware that our documentation advertizes it's use though.

> After all `update-bootloader` works great on Debian because there are many update-XY tools, some upstream and some from Debian.

Fully agreed. "`update-XYZ`" is great. "`mkinitrd`" was a fair name, but "`dracut`" - no comment.

mwilck:
Thanks for your review!

regenerate-initrd-posttrans
```diff
@@ -55,8 +48,7 @@ for f in "$dir"/*; do
 		echo "$0: /boot/$image does not exist, initrd won't be generated"
 		continue
 	fi
-	run_mkinitrd_setup
-	if ! /sbin/mkinitrd -k "/boot/$image" -i "/boot/$initrd"; then
+	if ! "$DRACUT" -f "/boot/$initrd"; then
```
mwilck, line 51: It turns out that this is wrong. I was under the impression that dracut would "guess" the kernel from the initrd file name, but it doesn't. If the kernel version isn't specified, it will generate the initrd for the *currently running kernel*, which is of course very wrong here.

weak-modules2
```diff
@@ -380,15 +382,8 @@ run_depmod_and_mkinitrd() {
 		mkdir -p /run/regenerate-initrd
 		touch /run/regenerate-initrd/$image-$krel
 	    else
-		doit /sbin/mkinitrd -k /boot/$image-$krel -i /boot/initrd-$krel
+		doit "$DRACUT" -f /boot/initrd-$krel
```
mwilck, line 385: This is wrong, too.

regenerate-initrd-posttrans
mwilck: Solved in 16.0.2

weak-modules2
mwilck: Solved in 16.0.2 (89030b5, fc1c54a)

hramrach: you can also call `dracut -f --kver $krel`

mwilck: I know. I decided to use the more conventional syntax. It really doesn't matter. Only **not** specifying the kernel version at all, as I did in this PR, was wrong.

mwilck: Luckily the broken code hadn't been submitted to Factory yet.

PR #22, opened 2021-05-24
Unify ppc64 and ppc64le

hramrach:


mwilck:
Do we need ppc64 at all?

hramrach:
In Tumbleweed

mwilck:
Hm, just saw that TW even still has "ppc" ... does that mean I have to undo the removal of `modprobe.conf.ppc` ?

hramrach:
And is that an actual release or is it something we build to have 32bit libraries?

mwilck:
I don't know, I'm asking you :grin:
I just saw that there are repositiories with arch "ppc" for openSUSE:Factory:PowerPC.

hramrach:
I see only 64bit releases http://download.opensuse.org/ports/ppc/tumbleweed/iso/

mwilck:
So, for my understanding - the `ppc` arch is really only a compatibility layer to allow execution of 32bit binaries on `ppc64{,le}`?

hramrach:
on pp64 only. there is no 32bit compatibility on ppc64le

mwilck:
Thanks. So, no need to resurrect `ppc`. Thanks for the clarification.

PR #25, opened 2021-05-28
Stop installing configuration files into /etc

mwilck:
See commit descriptions for rationale of implentation details.

Discussion revealed that (for now, at least) the best location for the `modprobe.d` and `depmod.d` directories is `/lib`, not `/usr/lib`, because *kmod* doesn't fully support `/usr/lib` yet and the usr merge will move them transparently to `/usr/lib`.

I tested the behavior on update of modified configuration files. From the past filesystem blacklisting endeavour, `60-blacklist_fs-$FS.conf` will be treated differently depending on 3 different factors:

 1) if the existing file under `/etc/modprobe.d` exists and is marked unmodifyable by the user (`__THIS FILE MAY BE MODIFIED_` marker removed), it's left untouched.
 2) otherwise, if the fs module is loaded at the time the postinstall script runs, a new `/etc/modprobe.d` file is created in which the blacklist directive is commented out.
 3) otherwise, if the file is user-modified, it is left untouched.
 4) otherwise, it is wiped from `/etc/modprobe.d`.

I've tested that behavior for all the above cases.

I've also tested that `modprobe.d` and `depmod.d` files are read correctly by *modprobe* and *depmod*, respectively.


hramrach:
An this is exactly the problem with doing nothing about kmod not reading /usr/lib.

If we want to get rid of files in /lib we will have to go through this dance twice, we will set bad examples for file location, etc.

hramrach:
Actually our current modprobe fully supports /usr/lib but not /lib

mwilck:
> An this is exactly the problem with doing nothing about kmod not reading /usr/lib.

I am not against changing kmod. It's just none of my business. I need to make sure *suse-module-tools* works correctly with kmod either way. The only thing that I'm interested in is that I must *know* which locations kmod supports in order to avoid installing configuration files in paths where kmod will ignore them.

> If we want to get rid of files in /lib we will have to go through this dance twice, we will set bad examples for file location, etc.

In SLE, and TW before `/usr` merge, `/lib/modprobe.d` is a perfectly valid location. In Factory after usr merge, it's identical to `/usr/lib`, so why bother? Moving from `/lib` to `/usr/lib` later will be much easier than the current move. The complication this time is not the location move itself. It is the fact that files under `/etc` are supposed to be modified, and files under either `/lib` or `/usr/lib` aren't.

> Actually our current modprobe fully supports /usr/lib but not /lib

which is a change that *you* made *after* the discussion about `suse-module-tools` had settled upon `/lib`. We should discuss this elsewhere.

hramrach:
> -------------------------------------------------------------------
> Fri Nov  6 11:41:50 UTC 2020 - Jan Engelhardt <jengelh@inai.de>
> 
> - Add usr-lib-modprobe.patch [boo#1092648]

This is where support for /usr/lib was added (before update to v28)

before that modprobe.d was read from /lib only and depmod.d from /usr/lib only.

mwilck:
Dependent packages like `suse-module-tools` can't rely on changelog entries. They must be able to figure out which search locations `kmod` supports, **from rpm dependencies**. We must **avoid hard** to install configuration files in paths where they'll be ignored.

In this context, it doesn't matter much if you add a new search path. But removing a search path matters a lot.

hramrach:
And by rpm dependencies kmod 28 and later supports /usr/lib on (open)SUSE for modprobe.d

There is no intention to remove support for directories. The question that needs to be answered is if there could be problem reading a directory twice under usrmerge and one should be dropped from the kmod search path or if reading both is fine even under usrmerge. That is only a technical detail of how kmod locates the files, in any case both /usr/lib and /lib should remain supported.

lnussel:
https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/tree/libkmod/libkmod-config.c#n808 ignores duplicates so no harm done reading both /lib and /usr/lib paths

suse-module-tools.spec
```diff
@@ -192,44 +180,88 @@ test_allow_on_install()
...
+    moved=
+    for f in %{_sysconfdir}/modprobe.d/*.conf.rpmsave %{_sysconfdir}/depmod.d/*.conf.rpmsave; do
+	case "$f" in *\*.conf.rpmsave) continue;; esac
+	mv -fv "$f" "${f%.rpmsave}"
+	moved=yes
+    done
+
+    if [ x$moved = xyes ]; then
+	echo "Local configuration changes restored after %{name} update." >&2
+	sleep 2
```
lnussel, line 260: please no!

suse-module-tools.spec
```diff
@@ -192,44 +180,88 @@ test_allow_on_install()
...
+check_lib_move() {
+    local pkg f
+
+    rm -f "/run/%{name}-lib-move"
+
+    # 99-local.conf was dropped with the /lib move
+    [ -f %{_sysconfdir}/modprobe.d/99-local.conf ] || return 0
+
+    # check if 99-local.conf is owned by suse-module-tools
+    pkg=$(rpm -qf --qf "%{name}" %{_sysconfdir}/modprobe.d/99-local.conf 2>/dev/null)
```
lnussel, line 199: I wouldn't do this in a scriptlet

lnussel (changes requested):
i cannot really review this. The scriptlets are ugly already. I'd really have to invest more time to figure out what this all is trying to achieve and how to improve it.
All this blacklisting black magic seems to be a SLE thing due to support anyway. How about just dropping that?

mwilck:
> All this blacklisting black magic seems to be a SLE thing due to support anyway. How about just dropping that?

No, it was deliberately made for openSUSE as well, for security reasons. It was discussed int the factory-ML thread was "[PLEASE SPEAK UP] Disabling legacy file systems by default?" (Jan, 2019). Actually the list of blacklisted filesystems is  much shorter on SLE, because we're shipping far less "exotic" FS drivers there. See also jsc#SLE-4085.

I considered removing the auto-unblacklisting code, because it's relatively safe to assume that noone will update to this version from a version < 15.1.10 / 15.2.0. There's no guarantee though, AFAICT.

suse-module-tools.spec
mwilck: ok, np

suse-module-tools.spec
mwilck: What harm does it do? Can you tell me some other way to figure out if the file, if it exists, belongs to the package being updated?

The point is that I want to avoid activating the "rename `.conf.rpmsave` to `.conf` code in `%posttrans` in all cases, except if I'm updating from the version that still holds config files in `/etc` to a version that has moved them to `/lib`. I think that's important.

mwilck:
> No, it was deliberately made for openSUSE as well, for security reasons.

And interestingly, I didn't see a single bug report against this. So either nobody uses these file systems in practice, or the auto-unblacklisting actually worked :-)

mwilck:
Tests were successful after appyling the above fixes.

mwilck:
@hramrach, @lnussel: sorry for the noise. I'm making mistakes ... will update again once I've fixed them.

suse-module-tools.spec
mwilck: Figured it out ... once the `/etc/modprobe.d` files have been dropped from the package, no more `.rpmsave` files will be generated for them. Thus my `rpm -qf --qf "%{name}` check is redundant.

I resolved this in the version just pushed.

mwilck:
@hramrach, @lnussel, I hope it's better now. Please have another look if you please.  Tests still pass. 

One improvement over the first version is that I don't rename any more configuration files (from `.conf.rpmsave` to `.conf`) that don't belong to my package. In the first version, that could have happened. Also, the code is slighly less convoluted now.

lnussel:
I still somehow dislike the gereral approach but this version certainly is much better :-)
The part about unblacklisting loaded file systems could be simplified

    while read fs x; do
          test "$fs" != nodev || continue
          test -e "/usr/lib/modprobe.d/60-blacklist_fs-$fs.conf" || continue
          ln -s /dev/null "/etc/modprobe.d/60-blacklist_fs-$fs.conf"
    done  < /proc/filesystems

suse-module-tools.spec
```diff
@@ -126,48 +141,25 @@ mkdir -p %{buildroot}%{_defaultlicensedir}
...
-# Delete obsolete unsupported-modules file from SLE11
-rm -f %{_sysconfdir}/modprobe.d/unsupported-modules
-%else
-# Logic for releases below CODE 15
-%if 0%{?is_opensuse} == 1
-allowed=1
-%else
-allowed=0
-%endif
+%if 0%{?sle_version} < 150000
```
lnussel, line 156: that means basically sle12 so drop the block completely?

mwilck:
> The part about unblacklisting loaded file systems could be simplified

I don't deny that it's over-complex, and I like your solution. But I'm kind of reluctant to change it now.
I suppose your remark "not my package" indicates that you'd be ok with that. One feature of the current implementation is that users can avoid automatic unblacklisting of filesystems even if they occasionally use them, by removing the `__THIS FILE MAY BE MODIFIED__` line. 

I suppose this feature is extremely rarely used, if at all. Yet I wouldn't want to remove it silently at this time. Also, I spent a lot of time testing the current auto-unblacklisting code and I know it's error-prone to change it.

mwilck:
I've overhauled this once more after discussing with @lnussel. The biggest change is that I don't check for file system modules in use in `%post` any more. This allows a major cleanup of the scriptlets, which look quite neat now. The reason is that the `.rpmsave` mechanism that I am using for previously modified config files doesn't work for config files modified *during the update procedure* (i.e. in `%post`). I could try hacking around this and making the scriptlets even more convoluted as they already were, or make a cut and decide that we simply don't do this any more. Ludwig and I agreed to go for the latter.

AFAICS, problems would only arise if **all** following conditions apply:

 1) user needs blacklisted file system `$FS` for booting the system
 2) user has never un-blacklisted `$FS`, neither manually nor
automatically via previous suse-module-tools updates (%post)
 3) user has never installed suse-module-tools 15.1.10 or 15.2.0 or
 later (iow, hasn't upgraded suse-module-tools since early 2019).

Instead of the `%post` automatism, we provide a new interactive un-blacklisting mechanism now (as suggested by @lnussel) which offers to un-blacklist a file system module if the user runs `modprobe $FS`. Details in [README.md](https://github.com/openSUSE/suse-module-tools/blob/no-more-etc/README.md).

mwilck:
My testing has gone well so far, and I need move forward. Any more comments?

lnussel:
I guess yast also loads modules eg when partitioning? Does that work still?

mwilck:
> I guess yast also loads modules eg when partitioning? Does that work still?

I'd expect that yast would just `modprobe minix`, not `modprobe fs-minix`. The `unblacklist` helper only takes action when invoked in a terminal. Otherwise it will quit, and `modprobe minix` will do the same thing as usual.

mwilck:
FTR, as this was delayed and the /usr merge completed a month ago, I have now changed the modprobe.d and depmod.d locations to /usr/lib for Tumbleweed. This is submitted to Fatcory as suse-module-tools 16.0.5.

PR #27, opened 2021-06-17
Use faster lsinitrd alternative

mwilck:
This can speed up weak-modules2 considerably. Even more so if combined with faster compression (e.g. zstd --fast).

### DIFF
--- a/lsinitrd-quick
+++ b/lsinitrd-quick
@@ -1,0 +1,92 @@
+#! /bin/bash
+
+trap 'echo error in $BASH_COMMAND >&2; exit 1' ERR
+
+MG_CPIO=$'\x71\xc7'
+MG_CPIO1=070701
+MG_GZ=$'\x1f\x8b'
+MG_GZ1=$'\x1f\x9e'
+MG_BZ="BZh"
+MG_XZ=$'\xfd7zXZ'
+MG_LZ4=$'\x02\x21'
+MG_LZO=$'\x89LZO'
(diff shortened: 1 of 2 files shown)


hramrach:
Shouldn't there be an error message when lsinitrd is not found saying where it was seached?

Currently if you don't have one it will give you the shell error, and currently it does not have to be in /usr/bin which may cause regression in some weird environments. Giving verbose error there would be great.

hramrach:
Like LSINITRD_OPTIONS="/bin/foo /bar/baz" ; for i in $LSINITRD_OPTIONS ; ... if ! [ -x $LSINITRD ] ; then echo no lsinitrd in $LSINITRD_OPTIONS ; exit 1 ; fi

mwilck:
> Giving verbose error there would be great.

OK. Also adapted the style to the `find_depmod()` function.

hramrach:
Looks good, thanks.

It is not exciting to have separate independent lsinitrd implementation but the quick one is 5x faster.

The standard one provides more features which comes at a cost.

mwilck:
> It is not exciting to have separate independent lsinitrd implementation but the quick one is 5x faster.

Right. It's not meant as a full replacement. Btw part of the problem seems to be that dracut's "skipcpio" program is badly implemented and painfully slow, which is probably better in latest dracut; see https://github.com/dracutdevs/dracut/pull/1126. But it's not only that, we don't need to look into the contents of the archive as the standard lsinitrd does.

tiwai:
Looks good to me, too.  Thanks!

PR #28, opened 2021-08-18
usrmerge: load kernel sysctl file from /usr (boo#1184804)

lnussel:


mwilck (commented):
Unless I'm mistaken, not even the latest kernel builds install the sysctl.conf file into `/usr/lib/modules` yet ([kernel-binary.spec.in](https://github.com/openSUSE/kernel-source/blob/17c8c26d49262888e4a0b5f66ee2aa734f465c90/rpm/kernel-binary.spec.in#L695)). So this is bound to fail unless we somehow coordinate the kernel and **s-m-t** releases for TW. 

AFAICS you have added `@USRMERGE@` code to the kernel spec file (`post.sh`) to symlink the sysctl file from `/usr/lib/modules/$(uname -r)/sysctl.conf` to `/boot/sysctl.conf-$(uname -r)`, so the legacy code should work for the time being, even when the installation location for the kernel's `sysctl.conf` file is moved. Why don't we push that kernel change now, once this is done in TW, merge this PR, and finally, remove the compat symlink code?

lnussel:
Yeah the kernel is struggling with staging a bit. we can wait merging this until the kernel is in.

mwilck:
I am wondering about that code I linked

```
sysctl_file=%buildroot/boot/sysctl.conf-%kernelrelease-%build_flavor
```

is still hard-coded. I fail to see how your USRMERGE code would work.

mwilck:
I'm pondering whether s-m-t could try to read the sysctl file from both locations for a transitional period, but I haven't figured it out yet.

lnussel:
it only makes sense to read from one location as the kernel's post script so far installs a compat symlink.

> I am wondering about that code I linked
> 
> ```
> sysctl_file=%buildroot/boot/sysctl.conf-%kernelrelease-%build_flavor
> ```
> 
> is still hard-coded. I fail to see how your USRMERGE code would work.

Due to the requirement that the kernel spec file needs to be buildable old distros the files are installed as always into the buildroot and the moved further down:
https://github.com/openSUSE/kernel-source/blob/17c8c26d49262888e4a0b5f66ee2aa734f465c90/rpm/kernel-binary.spec.in#L968

mwilck:
> Due to the requirement that the kernel spec file needs to be buildable old distros the files are installed as always into the buildroot and the moved further down:
> https://github.com/openSUSE/kernel-source/blob/17c8c26d49262888e4a0b5f66ee2aa734f465c90/rpm/kernel-binary.spec.in#L968

Ugh, subtle. I missed that (and I realize why :smiley:). Why install and move, rather than install into different locations for Leap and TW in the first place? Anyway, not my business.

## What about this:

`/usr/lib/systemd/system/systemd-sysctl.service.d/kernel.conf`
```
[Unit]
Requires=kernel-sysctl.service kernel-sysctl-legacy.service
```

`/usr/lib/systemd/system/kernel-sysctl.service`
```
[Unit]
Description=Apply sysctl settings for %v
DefaultDependencies=no
Conflicts=shutdown.target
Before=systemd-sysctl.service
After=systemd-modules-load.service
ConditionPathExists=/usr/lib/modules/%v/sysctl.conf

[Service]
Type=oneshot
ExecStart=-/usr/lib/systemd/systemd-sysctl /usr/lib/modules/%v/sysctl.conf
RemainAfterExit=yes

[Install]
RequiredBy=systemd-sysctl.service
```

`/usr/lib/systemd/system/kernel-sysctl-legacy.service`
```
[Unit]
Description=Apply sysctl settings for %v (legacy)
DefaultDependencies=no
Conflicts=shutdown.target
Before=systemd-sysctl.service
After=systemd-modules-load.service
ConditionPathExists=!/usr/lib/modules/%v/sysctl.conf
RequiresMountsFor=/boot

[Service]
Type=oneshot
ExecStart=-/usr/lib/systemd/systemd-sysctl /boot/sysctl.conf-%v
RemainAfterExit=yes

[Install]
RequiredBy=systemd-sysctl.service
```

This ensures that the settings are read from the new location if and only if the `sysctl.conf` file exists there, without erroring out or printing confusing messages if it doesn't. The only backside I can see is that we'd still have an implicit `RequiresMountsFor=/boot` as long as we have `/kernel-sysctl-legacy.service`. But that shouldn't do much harm AFAICS.

mwilck:
I like the idea of the separate service file pretty much. It's much more transparent than that hidden `ExecStartPre` statement. We should do this on Leap, too.

lnussel:
I agree that the separate service file would make things more obvious. We woudn't even need the drop-in if the new service file has `RequiredBy=systemd-sysctl.service`. `kernel-sysctl.service` would just have to be added to the presets.

I wouldn't introduce kernel-sysctl-legacy.service though. As soon as the kernel passes through staging that service is obsolete.

mwilck:
> `kernel-sysctl.service` would just have to be added to the presets.

Well, changing presets is notoriously hard, at least under SLE. That's why I avoided it.

> I wouldn't introduce kernel-sysctl-legacy.service though. As soon as the kernel passes through staging that service is obsolete.

Hm, if we don't do this, the sysctls won't be applied for "legacy" kernel packages. Note that kernel is special, lots of people tend to keep older kernels around. If a 5.13.x kernel with the new install location is going to be released soon, I believe we should have both services at least until TW reaches 5.15 (assuming that nobody needs to keep kernels lower than 2 minor versions from current).

mwilck:
I've created an alternative PR along the lines of my previous idea in #29.

lnussel:
> Hm, if we don't do this, the sysctls won't be applied for "legacy" kernel packages. Note that kernel is special, lots of people tend to keep older kernels around. If a 5.13.x kernel with the new install location is going to be released soon, I believe we should have both services at least until TW reaches 5.15 (assuming that nobody needs to keep kernels lower than 2 minor versions from current).

It's not like the sysctls do anything really crucial. So I think you are a bit over cautious here TBH :-) IMO lets just wait a week or two after the kernel is moved and then switch locations. Should be enough.

mwilck:
I agree I'm a bit scrupulous, but better safe than sorry :smile:

So unless you strongly object, I'll proceed with #29.

lnussel:
no strong objection :)

mwilck:
closing in favor of #29.

PR #29, opened 2021-08-18
sysctl kernel-uname_r.conf: separate service units, kernel usr merge

mwilck:
We used to pull in the kernel-specific sysctl settings via an ExecStartPre
directive in systemd-sysctl.service. Create a separate service instead,
which is pulled in by systemd-sysctl. This is more transparent, and allows
more fine-grained control.

In preparation of the kernel /usr merge (bsc#1184804), create two separate
services boot-sysctl. service and kernel-sysctl.service which work with kernel
packages pre- and post- usrmerge, respectively. Only one of them will be
actually started because of a Condition directive.

"boot-sysctl.service" can be dropped when we may assume that no users keep
old kernel packages around that install their sysctl settings under /boot.

### DIFF
--- a/50-kernel-uname_r.conf
+++ b/50-kernel-uname_r.conf
@@ -1,5 +1,0 @@
-[Unit]
-RequiresMountsFor=/boot
-
-[Service]
-ExecStartPre=-/usr/lib/systemd/systemd-sysctl /boot/sysctl.conf-%v
--- a/boot-sysctl.service
+++ b/boot-sysctl.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Apply Kernel Variables for %v from /boot
+DefaultDependencies=no
+Conflicts=shutdown.target
+Before=systemd-sysctl.service
+After=systemd-modules-load.service
+ConditionPathExists=!/usr/lib/modules/%v/sysctl.conf
+RequiresMountsFor=/boot
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/systemd/systemd-sysctl /boot/sysctl.conf-%v
+RemainAfterExit=yes
+
+[Install]
+WantedBy=systemd-sysctl.service
--- a/kernel-sysctl.service
+++ b/kernel-sysctl.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Apply Kernel Variables for %v
+DefaultDependencies=no
+Conflicts=shutdown.target
+Before=systemd-sysctl.service
+After=systemd-modules-load.service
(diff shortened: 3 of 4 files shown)


mwilck:
Tested this on SLE15-SP3 and TW. It seems to work well. I changed the service description to "Apply Kernel Variables for %v" in order to match the log messages from `systemd-sysctl.service`.

Log output on SLE/Leap (only `boot-sysctl.service` is installed):
```
Aug 19 09:17:34 bremer systemd[1]: Starting Apply Kernel Variables for kernel 5.3.18-4.g0fe9f47-default from /boot...
Aug 19 09:17:34 bremer systemd[1]: Finished Apply Kernel Variables for kernel 5.3.18-4.g0fe9f47-default from /boot.
Aug 19 09:17:34 bremer systemd[1]: Starting Apply Kernel Variables...
Aug 19 09:17:34 bremer systemd[1]: Finished Apply Kernel Variables.
```

Log output on TW (both services are installed, pre-usrmerge):
```
Aug 19 09:10:14 apollon.suse.de systemd[1]: Starting Apply sysctl settings for kernel 5.13.8-1-default from /boot...
Aug 19 09:10:14 apollon.suse.de systemd[1]: Condition check resulted in Apply sysctl settings for kernel 5.13.8-1-default being skipped.
Aug 19 09:10:14 apollon.suse.de systemd[1]: Finished Apply sysctl settings for kernel 5.13.8-1-default from /boot.
Aug 19 09:10:14 apollon.suse.de systemd[1]: Starting Apply Kernel Variables...
Aug 19 09:10:14 apollon.suse.de systemd[1]: Finished Apply Kernel Variables.
```

Log output on TW (both services installed, simulated post-usrmerge):
```
Aug 19 09:12:35 apollon.suse.de systemd[1]: Condition check resulted in Apply sysctl settings for kernel 5.13.8-1-default from /boot being skipped.
Aug 19 09:12:35 apollon.suse.de systemd[1]: Starting Apply sysctl settings for kernel 5.13.8-1-default...
Aug 19 09:12:35 apollon.suse.de systemd[1]: Finished Apply sysctl settings for kernel 5.13.8-1-default.
Aug 19 09:12:35 apollon.suse.de systemd[1]: Starting Apply Kernel Variables...
Aug 19 09:12:35 apollon.suse.de systemd[1]: Finished Apply Kernel Variables.
```

So we get these "Condition check" messages now, but I think they are actually useful.

mwilck:
Tested this successfully now. The installation test fails for SLE15-SP2 and older because they don't support `%service_del_postun_without_install`, but I refuse adding conditionals to the spec file for this special case.

PR #30, opened 2021-08-30
weak-modules2: dlog must not return error

lnussel:
It's sometimes called last in other function so a return value != 0
would make those other function fail


lnussel:
Another way is to avoid `true && ...` constructs and use reverse logic  `false || ...` instead :)

PR #31, opened 2021-08-30
weak-modules2: handle kernel modules in /usr/lib/modules

lnussel:
On the input side the script needs to handle /usr/lib/modules. When
checking in the file system it does not necessarily have to as /lib
link to /usr/lib. So as long as the script potentially has to deal with
both locations it's ok to poke around in /lib to keep the code simple.
As soon as we only need to handle /usr/lib all occurrences can be
replaced.

### DIFF
--- a/weak-modules
+++ b/weak-modules
@@ -95,7 +95,7 @@ read_modules_list() {
 	    module_krels[n]=$(krel_of_module "${modules[n]}")
 	else
 	    # Try to extract the kernel release from the path
-	    set -- "${modules[n]#/lib/modules/}"
+	    set -- "${modules[n]#*/lib/modules/}"
 	    module_krels[n]=${1%%/*}
 	fi
     done
@@ -215,10 +215,10 @@ if [ -n "$add_modules" ]; then
 		module=${modules[n]}
 		module_krel=${module_krels[n]}
 		case "$module" in
-		/lib/modules/$krel/*)
+		*/lib/modules/$krel/*)
 		    continue ;;
 		esac
-		subpath=${module#/lib/modules/$module_krel/updates}
+		subpath=${module#*/lib/modules/$module_krel/updates}
 		weak_module=/lib/modules/$krel/weak-updates/${subpath#/}
 		if [ -r "$weak_module" ]; then
 		    weak_krel=$(krel_of_module "$weak_module")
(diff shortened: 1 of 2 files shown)


lnussel:
pending testing from my side. Not sure I can finish it before vacation. So I filed it already in case some issue comes up and someone needs to pick this up.

weak-modules
```diff
@@ -95,7 +95,7 @@ read_modules_list() {
 	    module_krels[n]=$(krel_of_module "${modules[n]}")
 	else
 	    # Try to extract the kernel release from the path
-	    set -- "${modules[n]#/lib/modules/}"
+	    set -- "${modules[n]#*/lib/modules/}"
```
mwilck, line 98: Simply using the `*` wildcard doesn't cut it. If we really must change our code here, we have to check whether the modules for the given kernel reside in `/usr/lib`  or in `/lib` and act accordingly. But honestly I'd prefer to stick with `/lib`.

mwilck:
With "on the input side", you mean the list of modules fed into the script via stdin?

mwilck:
Looking more closely, I'd very much prefer to pretend all modules still lived under `/lib/modules`, relying on the fact that `/lib/modules` resolves to  `/usr/lib/modules`. All else gets really ugly in wm2. We would have to distinguish cases like

 a ) kernel not usrmerged, KMP neither: `/lib/$krel/weak-updates/$mod` -> `/lib/$other/updates/$mod`
 b ) kernel usrmerged, KMP not: `/usr/lib/$krel/weak-updates/$mod` -> `/lib/$other/updates/$mod`
 c ) kernel not usrmerged, KMP usrmerged: `/lib/$krel/weak-updates/$mod` -> `/usr/lib/$other/updates/$mod`
 d) both usrmerged: `/usr/lib/$krel/weak-updates/$mod` -> `/usr/lib/$other/updates/$mod`
 
Of course, systems can mix both kernels and KMP which are and aren't usrmerged. At least in wm2 we can't assume that that's never the case. It'd be **so** much easier and less error-prone not to have to make this distinction and use a) all the time. Note that the code uses `readlink` to determine whether a symlink points to the "right" target, which could easily fail if the version of wm2 that created the symlink disagreed about the policy wheter to use the target under `/lib` or `/usr/lib`.

mwilck:
I've created a simpler approach in #31. Please have a look.

weak-modules
lnussel: $ modules=(/usr/lib/modules/foo/bar.ko /usr/lib/modules/foo/baz.ko); n=1; echo ${modules[n]#*/lib/modules/}
    foo/baz.ko

lnussel:
if your minimal solution gets the job done then lets do it,yes. Right now the kmps are not usrmerged, that's what I'm trying to solve. II think it's absolutely ok to access through /ilb/modules. The script just needs to be aware that input might refer to /usr/lib (like from rpm -ql which is done somewhere IIRC)

lnussel:
close in favor of #31

PR #32, opened 2021-08-30
weak-modules2: accept modules under /usr/lib/modules on stdin

mwilck:
This logic relies on the assumption that /lib/modules always resolves
to /usr/lib/modules. wm2 continues to work with /lib/modules only,
internally. All symlinks created will point to files under /lib/modules.

Thus the only point in the code where we need to account for /usr is
when we read the list of modules from stdin, and there we simply strip
off the /usr prefix, knowing that the result will still resolve to the
same file.

### DIFF
--- a/weak-modules2
+++ b/weak-modules2
@@ -323,7 +323,7 @@ find_kmps() {
 	    continue
 	fi
 	rpm -ql --nodigest --nosignature "$kmp" \
-	    | grep -Ee '^/lib/modules/[^/]+/.+\.ko(\.[gx]z|\.zst)?$' \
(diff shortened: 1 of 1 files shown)


mwilck:
This is meant as a simper alternative to #31. Yet untested - we need KMPs installing into /usr for that.  AFAICS, the KMPs shipped with 5.13.12_2 still install under `/lib/modules`.

lnussel:
:+1:

lnussel:
any ETA for merging this so I can submit the change to move modules to the kernel package?

mwilck:
As you tested this and I don't see how it would do harm, merging.

PR #33, opened 2021-09-06
Import kernel scriptlets from kernel-source.

mwilck:

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,15 @@ Use the following environment variables:
    commands called by weak-modules2. Equivalent to --debug.
  * WM2_LOGFILE: redirect the output to the given file.
 
+## Kernel scriptlet files
+
+The kernel-rpm-script, kernel-kmp-script, kernel-inkmp-script, and
+kernel-cert-script are used internally by kernel packages.
+
+### Capturing log output from kernel scripts
+
+ * KERNEL_PACKAGE_SCRIPT_DEBUG when non-empty enables some extra output to kernel log.
+
 ## Kernel-specific sysctl settings
 
 This package installs the file `50-kernel-uname_r.conf` which makes sure
--- a/kernel-scriptlets/cert-script
+++ b/kernel-scriptlets/cert-script
@@ -1,0 +1,95 @@
+#!/bin/sh
+
+op=${0##*-}
+
+ca_check=""
+certs=""
+
+while true ; do
+    case $1 in
+	--ca-check)
+	    ca_check="$2"
+	    shift
+	    ;;
+	--certs)
+	    certs="$2"
+	    shift
+	    ;;
+	*) break
+	    ;;
+    esac
+    shift
+done
+
+run_mokutil () {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo mokutil "$@" >&2
+    mokutil "$@"
+}
+
+[ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo "cert $op" ca-check: "$ca_check" certs: "$certs" -- "$@" >&2
+
+script_rc=0
+
+case $op in
+    pre)
+	;;
+    post)
+	if command -v mokutil >/dev/null; then
+	    # Only apply CA check on the kernel package certs (bsc#1173115)
+	    if [ -n "$ca_check" ] && mokutil -h | grep -q "ca-check"; then
+		MOK_ARG="--ca-check"
+	    else
+		MOK_ARG=""
+	    fi
+	    # XXX: Only call mokutil if UEFI and shim are used
+	    for cert in $certs; do
+		cert="/etc/uefi/certs/${cert}.crt"
+		run_mokutil --import "$cert" --root-pw ${MOK_ARG}
+		rc=$?
+		if [ $rc != 0 ] ; then
+		    script_rc=$rc
+		    echo "Failed to import $cert" >&2
+		fi
+	    done
+	fi
+	;;
+    preun)
+	if command -v mokutil >/dev/null; then
+	    for cert in $certs; do
+		cert="/etc/uefi/certs/${cert}.crt"
+		# Here we queue the certificate for de-enrollment. If by postun
+		# the certificate does not exist last kernel using it was
+		# removed and we can queue it for de-enrollment with mokutil.
+		ln "$cert" "$cert.delete" || script_rc=$?
+	    done
+	fi
+	;;
+    postun)
+	if command -v mokutil >/dev/null; then
+	    for cert in $certs; do
+		cert="/etc/uefi/certs/${cert}.crt"
+		# If the certificate is now gone there are no users left
+		# queue -> de-enrollment
+		if ! test -e "$cert"; then
+		    run_mokutil --delete "$cert.delete" --root-pw
+		    rc=$?
+		    if [ $rc != 0 ] ; then
+			script_rc=$rc
+			echo "Failed to delete $cert" >&2
+		    fi
+		fi
+		rm "$cert.delete" || script_rc=$?
+	    done
+	fi
+	;;
+    posttrans)
+	;;
+    *)
+	echo Unknown scriptlet "$op" >&2
+	script_rc=255
+	;;
+esac
+
+exit $script_rc
+
+# vim: set sts=4 sw=4 ts=8 noet:
--- a/kernel-scriptlets/inkmp-script
+++ b/kernel-scriptlets/inkmp-script
@@ -1,0 +1,128 @@
+#!/bin/bash
+
+op=${0##*-}
+
+name=""
+version=""
+release=""
+kernelrelease=""
+flavor=""
+variant=""
+usrmerged="0"
+image=""
+certs=""
+
+while true ; do
+    case $1 in
+	--name)
+	    name="$2"
+	    shift
+	    ;;
+	--version)
+	    version="$2"
+	    shift
+	    ;;
+	--release)
+	    release="$2"
+	    shift
+	    ;;
+
+	--kernelrelease)
+	    kernelrelease="$2"
+	    shift
+	    ;;
+	--flavor)
+	    flavor="$2"
+	    shift
+	    ;;
+	--variant)
+	    variant="$2"
+	    shift
+	    ;;
+
+	--usrmerged)
+	    usrmerged="$2"
+	    shift
+	    ;;
+	--image)
+	    image="$2"
+	    shift
+	    ;;
+	--certs)
+	    certs="$2"
+	    shift
+	    ;;
+
+	*) break
+	    ;;
+    esac
+    shift
+done
+
+wm2=/usr/lib/module-init-tools/weak-modules2
+nvr="$name"-"$version"-"$release"
+
+if [ "$usrmerged" -ne 0 ] ; then
+    modules_dir=/usr/lib/modules/$kernelrelease-$flavor
+    system_map=${modules_dir}/System.map
+else
+    modules_dir=/lib/modules/$kernelrelease-$flavor
+    system_map=/boot/System.map-$kernelrelease-$flavor
+fi
+
+run_wm2() {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo wm2 "$@" >&2
+    /bin/bash -${-/e/} $wm2 "$@"
+}
+
+[ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || \
+    echo KMP "$op" name: "$name" version: "$version" release: "$release" \
+    kernelrelease: "$kernelrelease" flavor: "$flavor" variant: "$variant" \
+    usrmerged: "$usrmerged" image: "$image" certs: "$certs" -- "$@" >&2
+
+script_rc=0
+
+case $op in
+    pre)
+	;;
+    post)
+	if [ -x "$wm2" ]; then
+	    rpm -ql "$nvr" | INITRD_IN_POSTTRANS=1 run_wm2 --add-kernel-modules $kernelrelease-$flavor || script_rc=$?
+	fi
+	;;
+    preun)
+	rpm -ql "$nvr" | sed -n '/\.ko\(\.xz\|\.gz\|\.zst\)\?$/p' > "/var/run/rpm-$nvr-modules" || script_rc=$?
+	;;
+    postun)
+	if [ -e "$system_map" ]; then
+	    # the same package was reinstalled or just rebuilt, otherwise the files
+	    # would have been deleted by now
+	    # do not remove anything in this case (bnc#533766)
+	    rm -f "/var/run/rpm-$nvr-modules"
+	    exit 0
+	fi
+
+	modules=( $(cat "/var/run/rpm-$nvr-modules") )
+	rm -f "/var/run/rpm-$nvr-modules"
+	if [ ${#modules[*]} = 0 ]; then
+	    echo "WARNING: $nvr does not contain any kernel modules" >&2
+	    exit 0
+	fi
+	if [ -x "$wm2" ]; then
+	    printf '%s\n' "${modules[@]}" | run_wm2 --remove-kernel-modules $kernelrelease-$flavor || script_rc=$?
+	fi
+	;;
+    posttrans)
+	if test -x /usr/lib/module-init-tools/regenerate-initrd-posttrans; then
+	    /bin/bash -c 'set +e; /usr/lib/module-init-tools/regenerate-initrd-posttrans' || script_rc=$?
+	fi
+	;;
+    *)
+	echo Unknown scriptlet "$op" >&2
+	exit 255
+	;;
+esac
+
+exit $script_rc
+
+# vim: set sts=4 sw=4 ts=8 noet:
--- a/kernel-scriptlets/kmp-script
+++ b/kernel-scriptlets/kmp-script
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+op=${0##*-}
+
+name=""
+version=""
+release=""
+kernelrelease=""
+flavor=""
+usrmerged=""
+
+while true ; do
+    case $1 in
+	--name)
+	    name="$2"
+	    shift
+	    ;;
+	--version)
+	    version="$2"
+	    shift
+	    ;;
+	--release)
+	    release="$2"
+	    shift
+	    ;;
+
+	--kernelrelease)
+	    kernelrelease="$2"
+	    shift
+	    ;;
+	--flavor)
+	    flavor="$2"
+	    shift
+	    ;;
+	--usrmerged)
+	    usrmerged="$2"
+	    shift
+	    ;;
+
+	*) break
+	    ;;
+    esac
+    shift
+done
+
+wm2=/usr/lib/module-init-tools/weak-modules2
+nvr="$name"-"$version"-"$release"
+
+run_wm2() {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo wm2 "$@" >&2
+    /bin/bash -${-/e/} $wm2 "$@"
+}
+
+[ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || \
+    echo KMP "$op" name: "$name" version: "$version" release: "$release" \
+    kernelrelease: "$kernelrelease" flavor: "$flavor" \ usrmerged: "$usrmerged" -- "$@" >&2
+
+script_rc=0
+
+case $op in
+    pre)
+	;;
+    post)
+	if [ -x $wm2 ]; then
+	    INITRD_IN_POSTTRANS=1 run_wm2 --add-kmp $nvr || script_rc=$?
+	fi
+	;;
+    preun)
+	rpm -ql "$nvr" | sed -n '/\.ko\(\.xz\|\.gz\|\.zst\)\?$/p' > "/var/run/rpm-$nvr-modules" || script_rc=$?
+	;;
+    postun)
+	modules=( $(cat "/var/run/rpm-$nvr-modules") )
+	rm -f "/var/run/rpm-$nvr-modules"
+	if [ ${#modules[*]} = 0 ]; then
+	    echo "WARNING: $nvr does not contain any kernel modules" >&2
+	    exit 0
+	fi
+	if [ -x $wm2 ]; then
+	    printf '%s\n' "${modules[@]}" | run_wm2 --remove-kmp "$nvr" || script_rc=$?
+	fi
+	;;
+    posttrans)
+	if test -x /usr/lib/module-init-tools/regenerate-initrd-posttrans; then
+	    /bin/bash -c 'set +e; /usr/lib/module-init-tools/regenerate-initrd-posttrans' || script_rc=$?
+	fi
+	;;
+    *)
+	echo Unknown scriptlet "$op" >&2
+	exit 255
+	;;
+esac
+
+exit $script_rc
+
+# vim: set sts=4 sw=4 ts=8 noet:
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -1,0 +1,293 @@
+#!/bin/sh
+
+op=${0##*-}
+
+name=""
+version=""
+release=""
+kernelrelease=""
+flavor=""
+variant=""
+usrmerged="0"
+image=""
+certs=""
+
+while true ; do
+    case $1 in
+	--name)
+	    name="$2"
+	    shift
+	    ;;
+	--version)
+	    version="$2"
+	    shift
+	    ;;
+	--release)
+	    release="$2"
+	    shift
+	    ;;
+
+	--kernelrelease)
+	    kernelrelease="$2"
+	    shift
+	    ;;
+	--flavor)
+	    flavor="$2"
+	    shift
+	    ;;
+	--variant)
+	    variant="$2"
+	    shift
+	    ;;
+
+	--usrmerged)
+	    usrmerged="$2"
+	    shift
+	    ;;
+	--image)
+	    image="$2"
+	    shift
+	    ;;
+	--certs)
+	    certs="$2"
+	    shift
+	    ;;
+
+	*) break
+	    ;;
+    esac
+    shift
+done
+
+wm2=/usr/lib/module-init-tools/weak-modules2
+nvr="$name"-"$version"-"$release"
+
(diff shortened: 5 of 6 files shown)


kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+mok_rc=0
+
+case $op in
+    pre)
+	;;
+    preun)
+	if command -v mokutil >/dev/null; then
+	    for cert in "$certs"; do
+		cert="/etc/uefi/certs/${cert}.crt"
+		ln "$cert" "$cert.delete" || mok_rc=$?
```
mwilck, line 36: Please add a comment here ("preserve cert so that we can run in through mokutil in postun").

Btw why are we doing it like this? What's wrong with calling `mokutil --delete` in `%pre` ?

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+		    if [ $rc != 0 ] ; then
+			mok_rc=$rc
+			echo "Failed to delete $cert"
+		    fi
+		fi
+		rm "$cert.delete" || mok_rc=$?
+	    done
+	fi
+	;;
+    posttrans)
```
mwilck, line 76: Rather than looking at `$certs` in `%postun`, we might as well iterate over all `*.delete` files in `%posttrans`.

kernel-inkmp-script
```diff
@@ -0,0 +1,97 @@
+#!/bin/sh
+
+op=$(echo "$0" | sed -e "s/.*-//")
+
+while true ; do
+    case $1 in
+	--version)
+	    version="$2"
+	    shift 2
```
mwilck, line 9: same remark about `shift` as above.

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+    case $1 in
+	--kmp)
+	    kmp="$2"
+	    shift 2
+	    ;;
+	--certs)
+	    certs="$2"
+	    shift 2
+	    ;;
+	*) break
```
mwilck, line 15: Why not error out if called with invalid arguments? I see in your kernel-source patch that you call the script with 

    /usr/lib/module-init-tools/kernel-cert-pre --kmp 1 --certs "@CERTS@" "$@"

from the kernel scriptlets. This means that the script would get passed  rpm the scriptlet argument. It should probably only do something if this argument was 0 (package deinstallation). Right?

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+		MOK_ARG=""
+	    fi
+	    # XXX: Only call mokutil if UEFI and shim are used
+	    for cert in "$certs"; do
+		cert="/etc/uefi/certs/${cert}.crt"
+		run_mokutil --import "$cert" --root-pw ${MOK_ARG}
+		rc=$?
+		if [ $rc != 0 ] ; then
+		    mok_rc=$rc
+		    echo "Failed to import $cert"
```
mwilck, line 55: please redirect to stderr

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+    postun)
+	if command -v mokutil >/dev/null; then
+	    for cert in "$certs"; do
+		cert="/etc/uefi/certs/${cert}.crt"
+		if ! test -e "$cert"; then
+		    run_mokutil --delete "$cert.delete" --root-pw
+		    rc=$?
+		    if [ $rc != 0 ] ; then
+			mok_rc=$rc
+			echo "Failed to delete $cert"
```
mwilck, line 69: again, `>&2` please

kernel-inkmp-script
```diff
@@ -0,0 +1,97 @@
...
+	    ;;
+    esac
+done
+
+wm2=/usr/lib/module-init-tools/weak-modules2
+nvr="$name"-"$version"-"$release"
+
+run_wm2() {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo wm2 "$@" > /dev/kmsg
+    /bin/bash -${-/e/} $wm2 "$@"
```
mwilck, line 59: again, please don't use `kmsg`

kernel-kmp-script
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
```
mwilck, line 1: I think we should integrate this code entirely into `weak-modules2`.

kernel-kmp-script
```diff
@@ -0,0 +1,85 @@
...
+
+op=$(echo "$0" | sed -e "s/.*-//")
```
mwilck, line 3: see above

kernel-kmp-script
```diff
@@ -0,0 +1,85 @@
...
+
+while true ; do
+    case $1 in
+	--version)
+	    version="$2"
+	    shift 2
```
mwilck, line 9: see above

kernel-kmp-script
```diff
@@ -0,0 +1,85 @@
...
+	    ;;
+    esac
+done
+
+wm2=/usr/lib/module-init-tools/weak-modules2
+nvr="$name"-"$version"-"$release"
+
+run_wm2() {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo wm2 "$@" > /dev/kmsg
+    /bin/bash -${-/e/} $wm2 "$@"
```
mwilck, line 43: see above (`kmsg`)

kernel-kmp-script
```diff
@@ -0,0 +1,85 @@
...
+	if [ ${#modules[*]} = 0 ]; then
+	    echo "WARNING: $nvr does not contain any kernel modules" >&2
+	    exit 0
+	fi
+	if [ -x $wm2 ]; then
+	    printf '%s\n' "${modules[@]}" | run_wm2 --remove-kmp $nvr || script_rc=$?
+	fi
+	;;
+    posttrans)
+	eval "$(rpm -E "%{?regenerate_initrd_posttrans}")" || script_rc=$?
```
mwilck, line 75: Ugh. Please let's just copy the code from the macro here. Is there any reason not to do that?

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
+#!/bin/sh
+
+op=$(echo "$0" | sed -e "s/.*-//")
+
```
mwilck, line 4: see above

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+while true ; do
+    case $1 in
+	--version)
+	    version="$2"
+	    shift 2
```
mwilck, line 9: see above

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+	*) break
+	    ;;
+    esac
+done
+
+wm2=/usr/lib/module-init-tools/weak-modules2
+nvr="$name"-"$version"-"$release"
+
+trigger_purge_kernels() {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo Triggering purge-kernels > /dev/kmsg
```
mwilck, line 67: see above wrt `kmsg`

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+
+[ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || \
+    echo "$op" name: "$name" version: "$version" release: "$release" \
+    kernelrelease: "$kernelrelease" flavor: "$flavor" variant: "$variant" \
+    usrmerged: "$usrmerged" image: "$image" systemmap: "$systemmap" \
+    modulesdir: "$modulesdir" basepackage: "$basepackage" certs: "$certs" > /dev/kmsg
+
+script_rc=0
+
+case $op in
```
mwilck, line 103: This case statement is far too long. Please write shell functions and make the case statement just call them.

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+	# this will double the required space below /boot
+	# remove the files from the old packages to make room for the new initrd
+	# rpm may complain about low disk space if /boot/vmlinux does not fit
+	if [ "$basepackage" = 1 -a "$YAST_IS_RUNNING" != "" ]; then
+	    mydf="$( POSIXLY_CORRECT=1 df -P /boot/ | awk '/^(\/|-[[:blank:]])/{ print $4}' )"
+	    if test "$mydf" != "" ; then
+		echo "Free diskspace below /boot: $mydf blocks"
+		# echo "512 byte blocks: $(( 2 * 1024 * 20 ))"
+		if test "$mydf" -lt  "40960" ; then
+		    echo "make room for new kernel '"$flavor"' because there are less than 20MB available."
```
mwilck, line 119: if we really still need this, we want to increase the size limit, because the intird is rather 40MB than 20MB these days.

But that can be done in a separate step. No need to worry about it now.

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+		fi
+	    fi
+	fi
+
+
+	# On AArch64 we switched from 64k PAGE_SIZE to 4k PAGE_SIZE. Unfortunately
+	# btrfs can only use file systems created with the same PAGE_SIZE. So we
+	# check if the user has any btrfs file systems mounted and refuse to install
+	# in that case.
+	if [ $( uname -m ) = aarch64 -a \
```
mwilck, line 132: I think we should put "hacks" like this into a "plugin" directory in a later step.

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+	--systemmap)
+	    systemmap="$2"
+	    shift 2
+	    ;;
+
+	--modulesdir)
+	    modulesdir="$2"
+	    shift 2
+	    ;;
+	--basepackage)
```
mwilck, line 50: This is tested in almost every step below. I think we should provide separate scripts for basepackages and other packages.

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+	# compat stuff for /boot.
+	# if /boot and /usr are not speparate partitions we can just link
+	# the kernel there to save space. Otherwise copy.
+	if mountpoint -q /boot || mountpoint -q /usr; then
+	    copy_or_link="cp -a"
+	else
+	    copy_or_link="ln -sf"
+	fi
+	# XXX: need to fix suse-module-tools for sysctl.conf and System.map
+	for x in "$image" sysctl.conf System.map; do
```
mwilck, line 198: see bug 1189879, we need a fix here. Should be done separately though (probably before this is moved into **s-m-t**.

kernel-rpm-script
```diff
@@ -0,0 +1,302 @@
...
+		"$flavor" \
+		"$kernelrelease"-"$flavor" \
+		"$image"-"$kernelrelease"-"$flavor" \
+		initrd-"$kernelrelease"-"$flavor"
+	fi
+    fi
+
+    [ -z "$certs" ] || [ "$basepackage" -ne 1 ] || /usr/lib/suse-module-tools/kernel-cert-$op --certs "$certs" "$@"
+    ;;
+posttrans)
```
mwilck, line 292: While we are at it, please let's add a call to `/usr/lib/module-init-tools/regenerate-initrd-posttrans` here. It won't do anything anyway unless previous scriptlets left markers under `/run/regenerate-initrd`. Actually I would wish to have this call added to the kernel scriptlets before moving them to **s-m-t**.

hramrach:
There are quite a few build failures.

You probably want https://github.com/hramrach/suse-module-tools/commit/79c02fa0b09248104cec53213e570bb4bf5146ba or stop using obscpio archives.

What's even the point of using them?

Next to nothing understands them, not even the OBS services.

There is a number of fine archive formats that general purpose tools understand.

How do you look into an obscpio archive?

There is no way documented.

mwilck (commented):
> There are quite a few build failures.

Yes, they are due to a bug in the `obs-service-obs_scm-common` service. Don't worry about it. It's fine as long as Factory and Leap 15.3 builds succeed (they do).

> You probably want [hramrach@79c02fa](https://github.com/hramrach/suse-module-tools/commit/79c02fa0b09248104cec53213e570bb4bf5146ba) or stop using obscpio archives.

Not necessary, as you can see the build succeeds in my base repo (home:mwilck:suse-module-tools/suse-module-tools).

> What's even the point of using them?

I've been told it's better for OBS (saves space), as OBS is able to look into CPIO archives and record only diffs. I really don't care much about it. Anyway it's not the reason for the build failures.

> How do you look into an obscpio archive?

It's just plain cpio AFAIK. I never tried. The archive simply contains what's in git.

mwilck:
I've fixed the OBS builds now by pulling in a new version of the OBS service . SLE15-SP2 and older can't be fixed, they lack the `%systemd_postun_with_restart` macro.

mwilck:
Ftr, the problem on SLE-15-SP4 is https://github.com/openSUSE/obs-service-tar_scm/pull/406.

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+
+[ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo "cert $op" kmp: "$kmp" certs: "$certs" > /dev/kmsg
+
+mok_rc=0
+
+case $op in
+    pre)
+	;;
+    preun)
+	if command -v mokutil >/dev/null; then
```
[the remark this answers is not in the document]
hramrach: So long as there is no clause that does something else.

Also do we want to adhere to bash-free syntax or just make the script depend on bash?

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
+#!/bin/sh
+
+op=$(echo "$0" | sed -e "s/.*-//")
```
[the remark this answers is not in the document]
hramrach: but that would be bashism again

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+	    shift 2
+	    ;;
+	*) break
+	    ;;
+    esac
+done
+
+run_mokutil () {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo mokutil "$@" > /dev/kmsg
+    mokutil "$@"
```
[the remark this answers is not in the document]
hramrach: The messages that go to /dev/stderr tend to scroll by and get lost unless the user takes measures in advance. the terminal scrollback is typically not long enough for a zypper run.

But if you are setting a variable you can as well set up logging of your zypper output so no strong opinion on this one.

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
+#!/bin/sh
+
+op=$(echo "$0" | sed -e "s/.*-//")
+
+while true ; do
+    case $1 in
+	--kmp)
+	    kmp="$2"
+	    shift 2
```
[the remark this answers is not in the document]
hramrach: Why split the shift in 2?

Only options with parameter are supported which renders shift 2 when the option is processed.

There is no need to shift when an option is not processed so there is no point to have shift outside of the case.

Why are kmp or certs supposed to be initialized to empty?

Or should all variables that come as options be initialized to empty in case they leak from the environment?

kernel-rpm-script
hramrach: Large part of the script does not depend on basepackage, though. That would get duplicated. Also this is what is in the kernel now, splitting it at this time is needlessly hairy.

kernel-rpm-script
hramrach: Sounds like this is separate, orthogonal issue.

kernel-rpm-script
hramrach: That's kind of pointless because we are just removing the scriptlets anyway.

There is no posttrans for the kernel but I added it here for completeness and for potentially moving the initrd invocation here in a separate bug/feature.

kernel-rpm-script
hramrach: I am not opposed to restructuring this later so long as the interface - API that is provided now is future-proof for such restructuring. In this patch I want to import something that is really close to what was removed from the kernel.

kernel-rpm-script
hramrach: but if you really want separate basepackage script is should be done now so that the API defines separate scripts

kernel-cert-script
```diff
@@ -0,0 +1,86 @@
...
+	;;
+    post)
+	if command -v mokutil >/dev/null; then
+	    # Only apply CA check on the kernel package certs (bsc#1173115)
+	    if [ -z "$kmp" ] && mokutil -h | grep -q "ca-check"; then
+		MOK_ARG="--ca-check"
+	    else
+		MOK_ARG=""
+	    fi
+	    # XXX: Only call mokutil if UEFI and shim are used
```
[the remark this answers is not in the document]
hramrach: ???

The part that github displays as context for this comment does not look relevant

kernel-cert-script
hramrach: The cert is shared among many kernels. All of them on most systems that don't have custom kernels nor PTFs.

So you link the certificate here to queue it for deletion, and then in postun you delete it if the certificate no longer exists - last kernel using it was removed.

kernel-cert-script
hramrach: Yes, that's also fine. It does not really save us anything, though.

The certificate is de-enrolled and then the flag file is deleted so it is not de-enrolled more than once, and de-enrolling is cheap anyway, it just moves the flag to a different place where it is read at boot time by some EFI application.

kernel-cert-script
hramrach: The original scriptlet can check the argument, and if it does the check is preserved.

Note also that the check is not reliable in general and with multiversion in particular. The number passed is the number of packages installed.

There can be stricter checking of the passed options that reports unknown options (and possibly skips them because options starting with -- are distinct from the numeric argument passed by rpm).

kernel-cert-script
hramrach: sounds reasonable

kernel-kmp-script
hramrach: So long as the interface used in here is sufficient the code can be moved between s-m-t scripts later.

kernel-kmp-script
hramrach: ditto

kernel-kmp-script
hramrach: The macro can be defined differently in different codestreams?

It's just what the script does currently in the kernel translated to shell.

kernel-rpm-script
hramrach: ugh, these are the scripts from the kernel

I expect that much of this functionality will be restructured in the future or abolished entirely but at first I want to be as close to 1:1 as is reasonably possible.

kernel-cert-script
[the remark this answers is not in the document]
hramrach: -v is documented by command (1p) so I suppose it is standardized

If somebody complains that their limited shell does not have it we can solve it when that happens.

Note that the kernel uses this exact code and nobody complained.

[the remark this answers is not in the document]
hramrach: There is the argument check that whines about wrong argument which should be executed regardless of having mokutil so it should not be put outside of the case.

It can be made into a function/variable to make changing the -v easier in the future but it's not a big wine because it still has to be tested in each branch of the case anyway.

kernel-cert-script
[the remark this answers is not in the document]
hramrach: Right, if you refer to 
 # XXX: Only call mokutil if UEFI and shim are used
then I have no opinion on that.

Should be probably handled in a separate bug and the implications of any possible check discussed to death.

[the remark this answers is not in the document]
hramrach: Actually, there is the problem that on arm64 you suddenly get from no shim to shim on SP update without any warning so this is really hairy to get right. Really deserves a separate bug.

kernel-cert-script
hramrach: Also see https://github.com/hramrach/suse-module-tools for latest and possibly greatest revision of this patch.

kernel-cert-script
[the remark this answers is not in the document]
mwilck: I checked for bashism because you started the script with `#! /bin/sh`. I'm fine with using `#! /bin/bash` because `weak-modules2` needs it anyway. You're right wrt POSIX, so `check_bashisms` is probably simply wrong here.

kernel-cert-script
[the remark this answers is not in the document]
mwilck: no, `checkbashisms` accepts it, and it's in https://pubs.opengroup.org/onlinepubs/9699919799/nframe.html

kernel-cert-script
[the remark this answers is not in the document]
mwilck: > Only options with parameter are supported which renders shift 2 when the option is processed.

Perhaps we'll add parameter-less options in the future? 

> There is no need to shift when an option is not processed so there is no point to have shift outside of the case.

Well, I'd appreciate if you could do me the favor to do it this way. Shifting the options outside the `case` statement is a common idiom that makes the code more fool-proof and easier to read, for me at least. But I won't reject the PR if you don't.
 
> Why are kmp or certs supposed to be initialized to empty?

To avoid them being initialized from random variables from the environment. It's good practice in shell scripting.

> Or should all variables that come as options be initialized to empty in case they leak from the environment?

That's the idea, yes. Either that, or use well defined environment variables with expressive names for their initialization.

kernel-cert-script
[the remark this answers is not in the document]
mwilck: Note that depending on logging settings, **journald** may suppress log messages from `kmsg` that don't originate from the kernel. `>&2` is almost always the right thing to do, it makes the messages end up where the calling process sends them. This could be a log file, the journal, a terminal, or perhaps `/dev/kmsg` or `/dev/null`.

kernel-rpm-script
mwilck: Right. No objections. I just typed the thoughts that went through my head when I read this.

kernel-rpm-script
mwilck: Double-checking the part that doesn't depend on `basepackage` ...

1.  the aarch64/ PAGE_SIZE issue in `%pre`. I'm quite positive that this check should really depend on `basepackage`. `-extra` depends on the base pkg anyway, so if the latter can't be installed, the former won't be installable either.
 2.  `trigger_purge_kernels` in `%post` - unsure why that would be run by a non-base package. Same for the remval of `/boot/do_purge_kernels` in `%postun`.
 3. the `$usrmerged` stuff in `%post` - definitely only necessary for the base package.
 4. bootloader installation in `%post` - again, I see no point running that for a non-base package. Even worse for `%postun` - why remove a bootloader entry for `kernel-default` just because `kernel-default-extra` is being removed?
 5. `run_wm2 --remove-kernel` in `%postun` - I'm quite surprised to see this being called for non-basepackages. This is ancient code from 2011 that has never been changed in `postun.sh` - weird. Looks like a bug to me.

kernel-rpm-script
mwilck: right, this comment is only about the ordering of fixes.

kernel-rpm-script
mwilck: Well, once it's in **s-m-t** I'll have it under my control :grin:

kernel-rpm-script
hramrach: Maybe we should use kernel-inkmp-script instead?

kernel-cert-script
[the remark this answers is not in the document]
hramrach: The arguments accepted are pretty much final.

If at later point a kernel that passes more arguments exists it does not remove the kernels that don't from the repositories.

kernel-rpm-script
mwilck: > Maybe we should use kernel-inkmp-script instead?

I'm not sure I understand this remark correctly. We have 3 types of packages currently, kernel basepackages, kernel addon packages (like `-extra`), and KMPs. Each type gets different treatment by the scripts. I think we should leave it at that.

kernel-cert-script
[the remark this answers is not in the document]
mwilck: As I said, it won't be a reason to reject this PR even if I still have other style preferences. I can deal with it :-)

kernel-rpm-script
hramrach: And the difference between kernel-default-extra and kernel-default-dlm-kmp is what exactly?

mwilck: Excellent question. None, I would say. If I say "KMP", I mean real 3rd-party KMPs that install into `updates/` (or `extra/`), and for which we create `weak-updates` symlinks.

mwilck: The point is the wm2 invocation:

 * type 1 (basepackage): `--add-kernel`
 * type 2 (kernel extra): `--add-kernel-modules`
 * type 3 (KMP): `--add-kmp`

and `--remove-...`, respectively.

hramrach: the difference is INITRD_IN_POSTTRANS=1 for kernel-default-dlm-kmp but not kernel-default-extra

mwilck: ... which will be subject to change.

mwilck: That difference between `-extra` and the "kmp" packages doesn't make any sense to me.

hramrach: Neither to me so I switched -extra to the inkmp script. I suppose that the difference is mostly due to historical reasons that make the to scripts completely separate in kernel-source.

mwilck: To be on the safe side, I think we should take some of the issues we encountered to the kernel ML. I'll prepare a post.

kernel-cert-script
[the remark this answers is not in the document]
mwilck: OK. Opened bug 1190261 lest we forget.

kernel-cert-script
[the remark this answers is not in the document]
mwilck: I'm still not happy with this. My suggestion was `>&2`, not `2>&1`. We should keep stdout free of log messages and error messagse.

kernel-kmp-script
mwilck: > The macro can be defined differently in different codestreams?

Most certainly not. We move this functionality from rpm scriptlets to regular scripts now. Such scripts shouldn't need to run `rpm -E`. If we ever get a code stream that installs `regenerate_initrd_posttrans` elsewhere, we can change the path directly in the code (note that `r-i-p` is part of *s-m-t* itself).

mwilck:
Looking pretty good now. The only thing that I still dislike is the `rpm -E` call.

mwilck (commented):
All my issues have been addressed. Thanks for your patience!

PR #35, opened 2021-09-10
add udev rules from system-tuning-common-SUSE

tblume:
this was discussed in jira#SLE-21032
afterwards we should drop system-tuning-common-SUSE


mwilck:
Is this decided already?

tblume:
> Is this decided already?

Franck and I agree, the decision from PM seems to be still outstanding.

mwilck:
as this mainly concerns openSUSE for now, I also raised the question on #discuss-opensuse in slack.

mwilck:
Note that whenever files are migrated between packages, they need to be *removed* in the old package first before they can be added in another; otherwise file conflicts will be raised by the bots. And we need `Obsoletes:` and `Provides:` tags.

PR #36, opened 2021-09-14
rpm-script: link config also into /boot (boo#1189879)

lnussel:


mwilck (changes requested):
There are still a lot of missing ghost files. What about the symvers and symtypes files?
I would prefer having symlinks for all of these for now, or at least stating explicitly why we don't

```
> rpm -Vv kernel-default | grep missing
missing   g /boot/config-5.14.0-1-default
missing   g /boot/initrd-5.14.0-1-default-kdump
missing   g /boot/symtypes-5.14.0-1-default.gz
missing   g /boot/symvers-5.14.0-1-default.gz
missing   g /boot/vmlinux-5.14.0-1-default
missing   g /boot/vmlinux-5.14.0-1-default.xz
missing   g /usr/lib/modules/5.14.0-1-default/vmlinux
```

(except the last perhaps, I think it should be dropped).

lnussel:
Well, the whole purpose of ghost files is that they are allowed to be missing :-) I assume bug 1190434 is about some of the other files. I have yet to look into whether reintroducing a link or changing location is the best fix. Meanwhile we leave them missing. If no other complaints come up they can be removed from the file listing in the kernel package.

mwilck:
Well I'm going to take this now. But I tend to re-add the other symlinks too unless you provide very strong reasons why we shouldn't.

PR #38, opened 2021-09-14
Import kernel scriptlets from kernel-source.

hramrach:


mwilck:
Let's wait some more until the factory PR has settled, ok?

mwilck:
Actually, this PR is probably unnecessary, as up to now I've merged factory into SLE15-SP4, just changing the version number.

hramrach:
The scriptlets are now in Factory.

PR #42, opened 2021-09-16
Sles15 sp2

hramrach:


hramrach:
is there some way to trigger the OBS workflow?

mwilck:
> is there some way to trigger the OBS workflow?

IIUC the target branch must contain a workflow definition already (this is for security reasons, PR authors shouldn't be able to create arbitrary workflows). The SLE branches don't have workflows yet, the entire feature is experimental. And I don't even have OBS repositories for SLE *s-m-t* (only IBS).

hramrach:
The Factory workflow built against everything anyway so it could be reused for SLE with minimal change.

mwilck:
It doesn't matter much, but you're missing 6596bae ("rpm-script: link config also into /boot (boo#1189879)")

> The Factory workflow built against everything anyway so it could be reused for SLE with minimal change.

Sure. I'd just need to add the `workflows.yml`. I just didn't yet.

PR #43, opened 2021-09-21
sles12-sp5 Import kernel scriptlets from kernel-source.

hramrach:
Done as loose files - not sure the SourceX listing handles directories

### DIFF
--- a/kernel-scriptlets/cert-script
+++ b/kernel-scriptlets/cert-script
@@ -21,9 +21,19 @@ while true ; do
     shift
 done
 
+is_efi () {
+    local msg rc=0
+
 # The below statement fails if mokutil isn't installed or UEFI is unsupported.
 # It doesn't fail if UEFI is available but secure boot is off.
-if ! mokutil --sb-state 2>/dev/null; then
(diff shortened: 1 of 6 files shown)


mwilck:
> Done as loose files - not sure the SourceX listing handles directories

I don't understand, please explain why you can't do this like in the other distros.

hramrach:
AFAICT the files are tracked as loose sources in the .spec file rather than a tarball.

mwilck:
Please hold on. I'll see if I can change the SLE12 spec file to match the SLE15 one. That's more work for me but better for merging these scripts.

hramrach:
There is already modprobe.conf directory packaged as archive so I copied that part

hramrach:
might be a hard sell to maintenance moving all files into a tarball in a MU but we can try

mwilck:
The entire change will be a hard sell. But this is just a rearrangement of the sources. We must separate these steps logically - first change the spec file to match SLE15, then add the kernel scriptlets. If we do that, we could prove to maintenance that the build result of the the 1 step is identical to the 2nd step.

PR #44, opened 2021-09-21
Sles15

hramrach:


mwilck:
I'm merging this because it's looking fine. But I'm not sure what we need it for. IIRC there's no LTSS for SLE15.

hramrach:
but things are built against the SLE15 repo.

Not sure if the s-m-t should be updated or the build setup.

mwilck:
Will the SLE15 repo receive *kernel* packages that depend on the scriptlets being in s-m-t?

hramrach:
not the repo but cve/linux-4.12 builds against it

mwilck:
AFAICS it would only matter if the *running* kernel in the build VM on OBS needed the scriptlets for installation. Will that happen?

mwilck:
If we need this, I think we want to discuss pushing the update only to OBS for this purpose and not to customers. But TBH I'd prefer if the LTSS kernels for ancient distros kept the scriptlets in the kernel spec file.

hramrach:
yes, the cve branch can be updated  to build against SP1

mwilck:
> yes, the cve branch can be updated to build against SP1

I'm not sure I understand... so we don't need this PR in official s-m-t packages for SLE15-GA for the time being, right?

hramrach:
yes, we can avoid the update in SLE15-GA.

Once suse-module-tools is updates for all codestreams that receive 4.12 based kernel updates we can merge the packaging change to 4.12.

SLE 15 is not receiving kernel updates so it will not need s-m-t update, and we can avoid building new kernels against it.

mwilck:
Thanks. By the same rationale, I think we can skip this for SLE12 SP4 and older. Right?
(as you correctly observed, there are no separate branches for SLE12 SPs except SP5).

hramrach:
SLE 12 SP3 and older. SLE12 SP4 is still getting kernel updates from 4.12 branch.

And the package for 12 SP4 comes from SP2

mwilck:
OK, I'll submit to SP4.

hramrach:
> I'm merging this because it's looking fine. But I'm not sure what we need it for. IIRC there's no LTSS for SLE15.

LTSS for SLE15 ends in Dec 2022 so we need kernel updates there and hence s-m-t update as well.

mwilck:
ok, I'm going to submit it, too.

mwilck:
I recommended to maintenance to keep it internal as long as no kernel update  is required.

hramrach:
> I recommended to maintenance to keep it internal as long as no kernel update is required.

To the contrary. The sooner the package is released the better so that we can start building kernel with the packaging updates merged.

mwilck:
You can build the kernel right now. Are any kernel MUs for SLE15(GA) on the roadmap?

hramrach:
No, you cannot without the updated suse-module-tools.

KOTD is built every day, MUs are released every few months so there will be a few until the end of support.

mwilck:
Building should be possible just fine. Are you talking about post-build-checks?

Anyway feel free to comment under https://build.suse.de/request/show/258193

hramrach:
install check is part of the build in OBS so if required packages are not present the build does not start.

mwilck:
You don't need to BuildRequire this version, do you?

hramrach:
no, but you need to Require it and OBS basically adds Requires to BuildRequires.

We had this complaint with 15 SP2, SP4 and Factory already.

PR #54, opened 2021-11-24
Split conf files (bsc#1193059)

mwilck:
This PR is to illustrate the idea of [bsc#1193059](https://bugzilla.suse.com/show_bug.cgi?id=1193059).


mwilck:
re-pushed without the version bump

PR #55, opened 2021-11-30
boot-sysctl: make sure file exists

lnussel:
In containers no kernel might be installed. So make sure the service
does not fail.


mwilck:
In the spec file, we've got this:

```
# boot_sysctl may be dropped on TW when we can assume that nobody keeps
# kernel packages around that store sysctl files under /boot
%bcond_without boot_sysctl
```

So perhaps it's simply time to remove this file?

lnussel:
fine with me :)

lnussel:
the change makes sense also on released distros though as custom kernel may not ship with a sysctl file

PR #56, opened 2021-12-15
Make regenerate-initrd-posttrans compatible with Dracut's UEFI mode (…

brjsp:
…unified kernel image)

Right now, we blindly install Dracut's output to `/boot` regardless of what is chosen in `dracut.conf`.
Turns out, Dracut is smart enough to install into the correct location (either `boot` or the EFI partition)
depending on whether it is producing an iniutrd or a unified kernel image.
This change makes us compatible with the user setting `uefi=yes` inside `dracut.conf`,
and gives us automatic preliminary systemd-boot support (inherited from dracut)
with no changes needed in `perl-Bootloader` or anywhere else.

### DIFF
--- a/regenerate-initrd-posttrans
+++ b/regenerate-initrd-posttrans
@@ -54,7 +54,7 @@ for f in "$dir"/*; do
 		echo "$0: /boot/$image does not exist, initrd won't be generated"
 		continue
 	fi
-	if ! "$DRACUT" -f "/boot/initrd-$kver" "$kver"; then
+	if ! "$DRACUT" -f --kver "$kver"; then
 		err=$?
 	fi
 done


mwilck:
This isn't wrong but I don't see the benefit. On non-UEFI, the destination will be `/boot/initd-$kver`, and I actually prefer putting this path here explicitly than to rely on dracut's internal logic (which is currently patched under SUSE).

With `--uefi`, dracut chooses the path `/boot/efi/EFI/Linux/linux-5.15.7-1-default-a0385656b74c9241b77c1bb6577a603b.efi` for the current TW kernel. This is not what any other openSUSE tools expect (I hear you claim that *perl-Bootloader* does - I haven't tested this but tbh, I have my doubts). We don't install kernel or initrd under `/boot/efi` currently. Unless we fundamentally rework the way we install (open)SUSE on UEFI systems, I'd say that this doesn't fit our distro design. and that `dracut --uefi` is thus currently unsupported.

@lnussel, maybe you can contribute deeper insight than I have. Are we going to install initrds into `/boot/efi` any time soon, or start using `systemd-boot`?

brjsp:
@mwilck You misunderstood. Right now, `--uefi` does not work at all, because your script overrides the dracut output path. With this patch, I can set `/etc/sysconfig/bootloader` to `none`, bypassing `perl-Bootloader` completely, and instead rely on the fact that dracut and systemd-boot are made by upstream to work with each other.

mwilck:
> @mwilck You misunderstood. Right now, `--uefi` does not work at all, because your script overrides the dracut output path. With this patch, I can set `/etc/sysconfig/bootloader` to `none`, bypassing `perl-Bootloader` completely, and instead rely on the fact that dracut and systemd-boot are made by upstream to work with each other.

Ok. Waiting for @lnussel's comment. My main concern is that **s-m-t** would now depend on some implicit **dracut** defaults. That feels fragile.

> My use case was setting up secure boot with encrypted boot partition, btw. At first, I tried to set up GRUB, but it crashed because it did not recognize a LUKS container.

OT: argh, don't do that. I tried a few months ago and it's a PITA, because grub2's crypto implementation is so inefficient. Decrypting /boot took about 20s. And it was worse if I made a  mistake while typing the passphrase. Google for "grub cryptomount slow".

brjsp:
> OT: argh, don't do that. I tried a few months ago and it's a PITA, because grub2's crypto implementation is so inefficient. Decrypting /boot took about 20s. And it was worse if I made a mistake while typing the passphrase. Google for "grub cryptomount slow".
OT: thanks for the warning! Protecting the initrd from tampering seems like a common use case to me, and right now I know of two solutions: either put it on a encrypted partition, or use a unified kernel image. Both <a href=https://github.com/julian-klode/sicherboot>Debian</a> and <a href=https://github.com/andreyv/sbupdate>Arch</a> (neither of which use Dracut) have scripts accomplishing the second one.

Unless there is some other way to achieve that, which does not involve dracut integration?

lnussel:
I can't really comment on whether the patch is correct without digging deeper. In general I think we should make sd-boot possible to skip the complexity of perl-Bootloader etc. Indeed for sd-boot the kernel and initrd will end up in the efi partition and /boot loses it's current purpose. See also https://lnussel.github.io/2020/12/16/fslayout/

So I'm in favor of changes that helps us get there :-)

Wrt authenticating the initrd it seems a TPM2 chip can be of help there, see http://0pointer.net/blog/unlocking-luks2-volumes-with-tpm2-fido2-pkcs11-security-hardware-on-systemd-248.html

So if you are exploring that field @brjsp your feedback is very welcome :-) Feel free to post your findings on eg the factory mailinglist.

mwilck:
@aafeijoo-suse, @tblume:  If we merge this, it will be solely your responsibility as SUSE dracut maintainers to make sure that the initrd ends up in the right place that is supported / expected by downstream tools such as perl-Bootloader. AFAICS, this is already the case, but only because of some SUSE-specific patches which you'd either need to push upstream or keep maintaining.

If this is ok for you, I'll merge it.

mwilck:
More OT: @lnussel, not sure if you're aware that using the same file system for the root FS and storing the initrd has distinct advantages in terms of performance and disk utilization, as pointed out by @ddiss elsewhere. By modifying the FS layout such that the intiramfs always ends up elsewhere, we'd give up these advantages before we even started to benefit from them.

lnussel:
yeah, but we also pay a price for that... grub2

ddiss:
...
> My use case was setting up secure boot with encrypted boot partition, btw. At first, I tried to set up GRUB, but it crashed because it did not recognize a LUKS container.

Please raise a bugzilla ticket for this - as Martin indicated, it's slow, but should still boot successfully.

ddiss:
> More OT: @lnussel, not sure if you're aware that using the same file system for the root FS and storing the initrd has distinct advantages in terms of performance and disk utilization, as pointed out by @ddiss elsewhere. By modifying the FS layout such that the intiramfs always ends up elsewhere, we'd give up these advantages before we even started to benefit from them.

COW reflinks for initramfs are nice, but my main concern is the rollback / boot from snapshot functionality. I'm okay with moving to EFI + systemd-boot, but IMO this should not be the default case if Btrfs rootfs (including /boot) is used.
The boot-from-snapshot functionality provided with openSUSE / SLE is still a (quite unique) killer feature which IMO really shouldn't be thrown away without careful consideration.

lnussel:
https://lists.freedesktop.org/archives/systemd-devel/2021-November/047077.html

mwilck:
> > Ok. Waiting for @lnussel's comment. My main concern is that **s-m-t** would now depend on some implicit **dracut** defaults. That feels fragile.
> 
> @mwilck It already does, since we already call `dracut -f --regenerate-all`. That command uses dracut's default location.

Right.

mwilck:
FTR,` --kver` has been supported since dracut-021 (2012).

PR #58, opened 2022-01-04
cdrom: Disable autoclose by default (boo#1165047).

hramrach:
Please apply to 15 SP4 as well.


mwilck (changes requested):
Please call the file `80-options-cdrom.conf`, and add a bug ref in the conf file itself.

PR #59, opened 2022-01-13
rpm-script: force copy kernel to /boot (boo#1194501)

lnussel:


hramrach:
LGTM

Thanks

mwilck (changes requested):
Please add a short commit description. Bug ref is nice but I'd like the changelog to be readable without having to refer to BZ.

lnussel:
updated

mwilck:
Let's wait for the results of the discussion on BZ before we cherry-pick this into all other branches and open a dozen SRs.

hramrach:
The main reason to merge this which is unfortunately not mentioned in the commit message is to really reinstall all files provided by a package when it is reinstalled.

This only matters for usrmerge anyway so it's only relevant for Factory.

Also BTF support which triggers the problem is only available since SP4.

PR #60, opened 2022-06-01
Fixes for bsc#1199873 and bsc#1200107

mwilck:
- weak-modules2: don't hardcode /boot for kernel-related files
- driver-check.sh: don't hardcode /boot for kernel-related files
- driver-check.sh: avoid false positive error messages


driver-check.sh
```diff
@@ -10,6 +10,22 @@ warnings=0
 trap 'rm -rf "$tmp"' EXIT
 tmp=$(mktemp -d)
 
+find_usrmerge_boot() {
+    local filename=$1
+    local kver=$2
+    local ext=${3:+."$3"}
+    local f
+
+    for f in "/boot/$filename-$kver$ext" "/usr/lib/modules/$kver/$filename$ext"
```
lnussel, line 19: maybe reverse the order to always prefer the real location? /boot might be compat symlink only

weak-modules2
```diff
@@ -91,6 +91,22 @@ find_depmod() {
     dlog "DEPMOD=$DEPMOD"
 }
 
+find_usrmerge_boot() {
+    local filename=$1
+    local kver=$2
+    local ext=${3:+."$3"}
+    local f
+
+    for f in "/boot/$filename-$kver$ext" "/usr/lib/modules/$kver/$filename$ext"
```
lnussel, line 100: dito

driver-check.sh
lnussel: no is ok, that's why I didn't flag it as "request changes". Still feel like it's slightly more correct to revere the order, even though academic in practice :-)

mwilck: ok, I changed it.

mwilck:
Sorry for the many force-pushes; I'm trying to get the OBS CI to work, without success...

PR #61, opened 2022-07-04
Avoid bashism in rpm-script (bsc#1195391)

dirkmueller:
The rpm-script is using a bashism here:

  bash -${-/e/}

in order to call another script via bash. the easiest way to fix that
is to use bash -$- +e. While that hides the dependency on bash,
weak-modules2 will be detected as bash script so we have it always
available.

### DIFF
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -82,7 +82,7 @@ disarm_purge_kernels() {
 
 run_wm2() {
     [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo wm2 "$@" >&2
-    /bin/bash -${-/e/} $wm2 "$@"
+    /bin/bash -$- +e $wm2 "$@"
 }
 
 message_install_bl () {


hramrach:
No idea either. It's some piece of ancient code. What are these flags supposed to do?

mwilck:
> No idea either. It's some piece of ancient code. What are these flags supposed to do?

The shell options are passed down to wm2. So if the scriptlet was run with "-x", wm2 would run with "-x", too. The man page says "`$-` expands to the current option flags as specified upon invocation, by the set builtin command, or those set by the shell itself (such  as  the  -i  option)".

`${-/e/}` would remove the option `-e` (aka "bail out on error") from the option string if it was set.
```
> set -e
> echo $-
ehimBHs
> echo ${-/e/}
himBHs
> set +e
> echo $-
himBHs
> echo ${-/e/}
himBHs
```

hramrach:
And the shell options are not passed from the rpm scriptlet to this script so it is lost anyway by now.

mwilck:
> And the shell options are not passed from the rpm scriptlet to this script so it is lost anyway by now.

I don't understand what you mean. With Dirk's patch, the flags *would* still be passed.

mwilck:
> @dirkmueller , am I understanding correctly that we must add these dependencies?

No. We require `kernel-rpm-scriptlets`, which is provided by `suse-module-tools`, which depends on `/bin/bash`.

hramrach:
but no flags are passed to the script: https://github.com/openSUSE/kernel-source/blob/packaging/rpm/kernel-binary.spec.in#L1113

mwilck:
> but no flags are passed to the script: https://github.com/openSUSE/kernel-source/blob/packaging/rpm/kernel-binary.spec.in#L1113

Now I'm getting it, thanks ;-)

mwilck:
I created #62 as a simpler alternative.

mwilck:
s-m-t 16.0.21 is now in Base:System. Feedback appreciated.

dirkmueller:
this alternative solution also works for me, with the downside that any `set -x` etc will not be passed to the subshell (which I can live with, but I guess that has been the original reason for the way the code was written.

PR #62, opened 2022-07-06
kernel-scriptlets: don't pass flags to weak-modules2

mwilck:
The invocation of weak-modules2 was a bashism. We don't need
to pass any shell flags anyway. So replace it by a simple
command execution.

Alternative PR for #61.
Fixes #61.


mwilck:
@hramrach , @dirkmueller : please comment

hramrach:
LGTM

Thanks

PR #64, opened 2022-09-07
Split kernel scriptlets into separate sub-package

lnussel:
Allows to replace them with noops for systems based on file triggers only

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -64,7 +64,9 @@ Requires:       rpm
 Requires(post): /usr/bin/grep
 Requires(post): /usr/bin/sed
 Requires(post): coreutils
+%if 0%{?suse_version} < 1550
 Provides:       suse-kernel-rpm-scriptlets = 0
+%endif
 Provides:       udev-extra-rules = 0.3.0
 Obsoletes:      udev-extra-rules < 0.3.0
 Provides:       system-tuning-common-SUSE = 0.3.0
@@ -98,6 +100,17 @@ Supplements:    dkms
 This package contains the legacy "weak-modules" script for kernel
 module package (KMP) support. It was replaced by "weak-modules2" in
 SLE 11 and later. It is still used by the DKMS module packaging framework.
+
+%if 0%{?suse_version} >= 1550
+%package scriptlets
+Summary:        Kernel rpm scriptlets
+Provides:       suse-kernel-rpm-scriptlets = 0
+Requires:       suse-module-tools = %{version}
+Provides:       suse-module-tools:/usr/lib/module-init-tools/kernel-scriptlets
+
+%description scriptlets
+Scripts called by the SUSE kernel packages on installation
+%endif
 
 %prep
 %setup -q
@@ -263,8 +276,12 @@ exit 0
 %{_rpmmacrodir}/macros.initrd
 %endif
 %{_bindir}/kmp-install
-/usr/lib/module-init-tools
-%exclude /usr/lib/module-init-tools/weak-modules
+%dir /usr/lib/module-init-tools
+/usr/lib/module-init-tools/driver-check.sh
+/usr/lib/module-init-tools/lsinitrd-quick
+/usr/lib/module-init-tools/regenerate-initrd-posttrans
+/usr/lib/module-init-tools/unblacklist
+/usr/lib/module-init-tools/weak-modules2
 %{_unitdir}/*.service
 %{_unitdir}/systemd-sysctl.service.d
 %{_modulesloaddir}
@@ -272,6 +289,12 @@ exit 0
 %ifarch ppc64 ppc64le
 /usr/lib/systemd/system-generators
 %endif
+#
+%if 0%{?suse_version} >= 1550
+%files scriptlets
+%endif
+/usr/lib/module-init-tools/kernel-scriptlets
+
 
 %files legacy
 %defattr(-,root,root)


mwilck:
Can you please provide some more background for what you are aiming at? How wold "systems based on file triggers only" implement this same functionality?

Adding @hramrach.

lnussel:
Instead of packages calling scripts one can have file triggers installed by other packages that do the same or something else. Eg when using system-boot instead of grub one can call kernel-install. I made some PoC for that at https://build.opensuse.org/package/show/home:lnussel:legacyfree/kernel-install

mwilck:
Forgive my ignorance, I don't understand how this is supposed to work.  The kernel packages themselves require `suse-kernel-rpm-scriptlets`, and call the scriptlets during `%post`, etc. You want to use systemd's **kernel-install** instead. Shouldn't we discuss that part first? Is kernel-install feature-wise equivalent to the SUSE scriptlets? Does it handle split kernel packages (`-extra` etc) and  KMPs correctly (possibly some other SUSE features that I may have forgotten)? Handling KMPs in particular is difficult, and changes in that logic are likely to cause regressions.

IMO the first step would be to *change* the kernel scriptlets such that they call kernel-install instead of using their own logic, where possible. That would give us an idea if kernel-install is still lacking functionality. Future file triggers in whatever package should then invoke the scriptlets rather than calling kernel-install directly. This way we could move our functionality into `kernel-install.d` step by step. Eventually the scriptlets will just be shallow wrappers around kernel-install, at which point we can get rid of them, maybe.

AFAICS dropping the scripts would be about the last step in this endeavor.
I am certainly misunderstanding and/or overlooking essential things, so please explain.

lnussel:
With this change I'm merely allowing to replace the package with one that does something else. In my example replaces the scriptlets with noops. Could of course also provide an implementation that calls kernel-install directly. The point here really is to allow replacing the scriptlets. As long as they are part of the suse-module-tools package that's not possible without also replacing all the rest. Seems to me like the `suse-kernel-rpm-scriptlets` provides was created exactly for that purpose.

suse-module-tools.spec
```diff
@@ -99,6 +101,16 @@ This package contains the legacy "weak-modules" script for kernel
 module package (KMP) support. It was replaced by "weak-modules2" in
 SLE 11 and later. It is still used by the DKMS module packaging framework.
 
+%if 0%{?suse_version} >= 1550
+%package -n suse-kernel-rpm-scriptlets
+Summary:        Traditional kernel scriptlets
+Requires:       suse-module-tools = %{version}
+Provides:       suse-module-tools:/usr/lib/module-init-tools/kernel-scriptlets
```
mwilck, line 108: What exactly does this syntax mean?

suse-module-tools.spec
```diff
@@ -64,7 +64,9 @@ Requires:       rpm
 Requires(post): /usr/bin/grep
 Requires(post): /usr/bin/sed
 Requires(post): coreutils
+%if 0%{?suse_version} < 1550
 Provides:       suse-kernel-rpm-scriptlets = 0
+%endif
```
mwilck, line 69: Should s-m-t require `suse-kernel-rpm-scriptlets`?

suse-module-tools.spec
```diff
@@ -99,6 +101,16 @@ This package contains the legacy "weak-modules" script for kernel
 module package (KMP) support. It was replaced by "weak-modules2" in
 SLE 11 and later. It is still used by the DKMS module packaging framework.
 
+%if 0%{?suse_version} >= 1550
+%package -n suse-kernel-rpm-scriptlets
```
mwilck, line 105: Perhaps you should choose a package name that differs from the resolvable name?

suse-module-tools.spec
```diff
@@ -99,6 +101,16 @@ This package contains the legacy "weak-modules" script for kernel
...
+Summary:        Traditional kernel scriptlets
+Requires:       suse-module-tools = %{version}
+Provides:       suse-module-tools:/usr/lib/module-init-tools/kernel-scriptlets
+
+%description -n suse-kernel-rpm-scriptlets
+Traditional kernel scriptlets
```
mwilck, line 111: "traditional" is too subjective in my opinion. Please change this to a neutral wording.

suse-module-tools.spec
mwilck: Well, I guess not. Those packages that need them require the scripts directly.

mwilck (commented):
This must be seen in the context of bsc#1202353. We want to add a depenency on dracut. IIUC this dependency would be in s-m-t, not in the scriptlets package.

mwilck:
I understand your goal now. While I believe that for now you might as well just fork s-m-t for your "legacy-free" project, the split might actually work. See my minor comments above.

suse-module-tools.spec
lnussel: it's a split-provides: https://en.opensuse.org/openSUSE:Package_dependencies#Splitting_off_a_sub-package

suse-module-tools.spec
lnussel: Any suggestions?

mwilck: According to your own choice of words, maybe `suse-legacy-kernel-rpm-scriptlets`? But that's a bit lengthy... 
I really don't care much. But I think it would be good to separate package name and resolvable, in particular because you intend to exchange the package for something else.

lnussel:
Updated

mwilck:
I've pushed this into Base:System with sr#1005598. Not forwarded to Factory just yet.

mwilck:
This change has caused regressions (see https://build.opensuse.org/request/show/1007228).
The problem is that lots of packages depend on suse-module-tools explicitly. In the case of the **virtualbox** package, s-m-t was pulled in indirectly via `kmod`'s dependency on s-m-t.

This affects OBS, mainly. On normal systems with kernel-default installed, the scriptlets package will be installed because it's pulled in via the kernel. But `kernel-obs-build` doesn't depend on `suse-kernel-rpm-scriptlets`.

I am going to revert this change for now, as it's blocking other fixes. We need to add at least a dependency `kernel-obs-build → suse-kernel-rpm-scriptlets` before this can be pushed to factory.

lnussel:
kmp's are supposed to pull in the scriptlets. kernel-obs-build itself does not need it it. See `/usr/lib/rpm/kernel-module-subpackage` . The affected packages probably don't use `%_kernel_module_package`?

mwilck:
It was a build issue, AFAIU. There's no bug. I only have the comments from @Vogtinator on the OBS request (see above). Sorry but I didn't bother to dig very deeply. I had a look at the buildinfo for virtualbox, and saw that `suse-kernel-rpm-scriptlets` aren't pulled in.

lnussel:
http://bugzilla.opensuse.org/show_bug.cgi?id=1205149

PR #65, opened 2022-09-08
mokutil shall only be called when efivarfs is writeable

KilianHanich:
On certain UEFI implementations, like U-Boot's, UEFI variables cannot be
changed at runtime.
As such the Linux kernel only supports a read-only efivarfs.
This patch adds a check so that an update doesn't automatically fail like
described in https://bugzilla.opensuse.org/show_bug.cgi?id=1201066

### DIFF
--- a/kernel-scriptlets/cert-script
+++ b/kernel-scriptlets/cert-script
@@ -73,7 +73,7 @@ case $op in
 		fi
 	    done
 	else
-	    test -d /sys/firmware/efi/efivars && echo "WARNING: cannot update MOK variables since /sys/firmware/efi/efivars is not writeable" 1>&2
+	    test ! -d /sys/firmware/efi/efivars || echo "WARNING: cannot update MOK variables since /sys/firmware/efi/efivars is not writeable" 1>&2
 	fi
 	;;
     preun)
--- a/kernel-scriptlets/cert-script
+++ b/kernel-scriptlets/cert-script
@@ -60,16 +60,19 @@ case $op in
 	if [ -n "$ca_check" ] && mokutil -h | grep -q "ignore-keyring"; then
 	    MOK_ARGS="${MOK_ARGS} --ignore-keyring"
 	fi
-	# XXX: Only call mokutil if UEFI and shim are used
-	for cert in $certs; do
-	    cert="/etc/uefi/certs/${cert}.crt"
-	    run_mokutil --import "$cert" --root-pw ${MOK_ARGS}
-	    rc=$?
-	    if [ $rc != 0 ] ; then
-		script_rc=$rc
-		echo "Failed to import $cert" >&2
-	    fi
-	done
+	# XXX: Only call mokutil if UEFI and shim are used and efivarfs is writeable
+	test -w /sys/firmware/efi/efivars
+	if [ $? == 0 ] ; then
+	    for cert in $certs; do
+	        cert="/etc/uefi/certs/${cert}.crt"
+	        run_mokutil --import "$cert" --root-pw ${MOK_ARGS}
+	        rc=$?
+	        if [ $rc != 0 ] ; then
+		    script_rc=$rc
+		    echo "Failed to import $cert" >&2
+		fi
+	    done
+	fi
 	;;
     preun)
 	for cert in $certs; do
--- a/kernel-scriptlets/cert-script
+++ b/kernel-scriptlets/cert-script
@@ -64,14 +64,16 @@ case $op in
 	test -w /sys/firmware/efi/efivars
 	if [ $? == 0 ] ; then
 	    for cert in $certs; do
-	        cert="/etc/uefi/certs/${cert}.crt"
-	        run_mokutil --import "$cert" --root-pw ${MOK_ARGS}
-	        rc=$?
-	        if [ $rc != 0 ] ; then
+		cert="/etc/uefi/certs/${cert}.crt"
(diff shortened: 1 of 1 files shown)


kernel-scriptlets/cert-script
```diff
@@ -60,16 +60,19 @@ case $op in
...
-	    cert="/etc/uefi/certs/${cert}.crt"
-	    run_mokutil --import "$cert" --root-pw ${MOK_ARGS}
-	    rc=$?
-	    if [ $rc != 0 ] ; then
-		script_rc=$rc
-		echo "Failed to import $cert" >&2
-	    fi
-	done
+	# XXX: Only call mokutil if UEFI and shim are used and efivarfs is writeable
+	mount | grep efivarfs | grep -E '\(.*,?ro,?.*\)'
```
lnussel, line 64: does `test -w /sys/firmware/efi/efivars` work instead?

KilianHanich: Yes, that works too. While it would also mean an hard-coded path, considering that it's `/sys`, I don't think that it should be a problem.

mwilck:
Perhaps we should test with `mokutil --test-key` whether a given key is already enrolled, and warn the user otherwise?

mwilck:
What's the policy for updating MOK keys on systems with read-only efivarfs?

KilianHanich:
From what I can tell, we don't compile U-Boot with Secure Boot support.
Besides the fact that you then need to generate PK, KEK, db and dbx manually and need to tell U-Boot in its shell about them (manually).

About MOK keys: It would be actually interesting where shim actually saves these keys. If it's UEFI variables, you would only be able to upgrade them via recompiling U-Boot (and similar bootloaders).
So, if nobody here knows, I will ask them where they store these.

mwilck:
The list of keys that shim checks are in boot time variables `MokList` and `MokListX`. There are a lot of runtime variables like `MokNew`, `MokDel` etc. See [MokManager.c](https://github.com/rhboot/shim/blob/main/MokManager.c).

mwilck:
I think we should at least try to print a warning when a to-be-installed kernel uses a certificate that's not enrolled, and can't be enrolled. The user will most likely not be able to boot this kernel without disabling secure boot.

KilianHanich:
I think that this should be enough.

kernel-scriptlets/cert-script
```diff
@@ -60,16 +60,21 @@ case $op in
...
+		cert="/etc/uefi/certs/${cert}.crt"
+		run_mokutil --import "$cert" --root-pw ${MOK_ARGS}
+		rc=$?
+		if [ $rc != 0 ] ; then
+		    script_rc=$rc
+		    echo "Failed to import $cert" >&2
+		fi
+	    done
+	else
+	    test -d /sys/firmware/efi/efivars && echo "WARNING: cannot update MOK variables since /sys/firmware/efi/efivars is not writeable" 1>&2
```
mwilck, line 76: You shouldn't use `&&` constructs in these scripts, because the compound statement might have an error exit status. Please write

    [ ! -d /sys/firmware/efi/efivars ] || echo ...

instead (any specific reason why you use `test`?

KilianHanich: Well, the line just says "if it's a directory, echo, otherwise, don't".

Also, your line is also a compound statement (it just uses `||` instead of `&&`) and is equivalent to `test ! -d /sys/firmware/efi/efivars || echo ...`. So, I don't really get the difference in meaning here.

As for why I use `test`:
- I find it more readable
- it's easier to find documentation about it in case you don't know what it means (and yes, I consider things like that important); seriously, try to search for something like that on a search engine or a manpage

mwilck: We use `[` everywhere else, it's POSIX and well-established shell programming practice. But I'm not religious on this matter. 

I am religious wrt `&&` though. 

> Also, your line is also a compound statement (it just uses || instead of &&) and is equivalent to `test ! -d /sys/firmware/efi/efivars || echo ...`

    test ! -d $DIR || echo blah
    if test -d $DIR; then echo blah; fi
    test -d $DIR && echo blah

These statements are logically equivalent, and would be equivalent in most programming languages. But in shell, there's a difference. The first 2 variants always have non-zero exit status. But the last one has non-zero exit status if `$DIR` doesn't exist. If the shell is run with the `-e` flag, the script would abort immediately. This is why `&&` constructs should be avoided.

KilianHanich: About `test`: it's also part of POSIX (https://pubs.opengroup.org/onlinepubs/9699919799/nframe.html > "Shell & Utilities" > "Utilities" > "test"). And well, as I said, I like programs (and scripts) to be accessible to beginners and as such things should be easy to find in case you don't know about them (and well, test is easier to find than the other construct).

I will change the one about `&&`, thanks for the explanation.

mwilck: > About test: it's also part of POSIX

Sure. And we use both `test` and `[` in this code, so although I clearly favor `[`, I said I am not religious on this matter.

mwilck (changes requested):
Thanks for your contribution, it's appreciated. But I'm still not happy with this PR.
I believe we need the same logic in `postun` as well. I'll post an alternative PR.

mwilck:
Alternative PR #66 submitted. Please check if it works for you.

PR #66, opened 2022-09-12
cert-script: skip cert handling if efivarfs is not writable

mwilck:
On certain UEFI implementations, like U-Boot's, UEFI variables cannot be
changed at runtime. As such the Linux kernel only supports a read-only efivarfs.
This patch adds a check so that an update doesn't automatically fail,
like described in boo#1201066.

Fixes #65.

### DIFF
--- a/kernel-scriptlets/cert-script
+++ b/kernel-scriptlets/cert-script
@@ -39,7 +39,12 @@ if ! is_efi; then
     exit 0
 fi
 if [ ! -w /sys/firmware/efi/efivars ]; then
-    echo "$0: efivarfs is not writable, skipping certificate handling" >&2
+    mkdir -p /run/suse-kernel-rpm-scriptlets
+    if [ -n "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || \
+	   [ ! -e /run/suse-kernel-rpm-scriptlets/cert-warning ]; then
+	echo "$0: efivarfs is not writable, skipping certificate handling" >&2
+    fi
+    touch /run/suse-kernel-rpm-scriptlets/cert-warning
     exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


KilianHanich:
Tested it and it fixes this part, so fine by me.

Although it seems that it's not the only place where I need to go (shim's post script also has a problem, but didn't notice so far because I didn't have an update related to it so far). But well, it's unrelated to here.

KilianHanich:
> Tested it and it fixes this part, so fine by me.
> 
> Although it seems that it's not the only place where I need to go (shim's post script also has a problem, but didn't notice so far because I didn't have an update related to it so far). But well, it's unrelated to here.

PS: Unlike before you get the warning multiple times now tho (3 times to be exact because it gets called 3 times).
While I don't personally care about this, I know some who don't.

mwilck:
Good point. Should be improved now.

KilianHanich:
Well, from my pov this is fine now.
So, I guess only hramrach now needs to review this.

hramrach (approved):
LGTM

Thanks

hramrach:
Regarding updating the certificates on platforms that don't have writable EFI variables - if somebody manages to get secure boot working on such platform they need to implement updating the certificates as well. If it needs some support in these scripts the support can be merged once such solution exists.

Generally I think it's unlikely to happen with u-boot. While u-boot has support for the secure boot verification it is platform-specific (and usually broken) how the boot chain up to u-boot is secured. The chance that somebody picks a platform with non-broken security and works out the details for it is not big.

lnussel:
Could add `mokutil --sb-state &>/dev/null` as condition for the warning. It's also possible to drop a text file into `/var/adm/update-messages` for zypper and even yast to display the message explicitly at the end of the installation. I don' know where the official documentation for that is but you can find an example in mutt.

mwilck:
Thanks for the hint! That's a valuable general possible improvement for s-m-t. I believe this wouldn't be only case where we should place a prominent warning.

I see it as an enhancement request though.

mwilck:
> Could add mokutil --sb-state &>/dev/null as condition for the warning.

That's already the case, as this test is done after the call to `is_efi()`.

PR #67, opened 2022-11-04
80-hotplug-cpu-mem.rules: restrict cpu rule to x86_64

mwilck:
This causes undesirable log messages on ppc64le (bsc#1204423).
It has never been requested for any arch, except for `x86_64` in
bsc#703100. So activate the rule for x86_64 only.


hramrach:
That looks nice

Thanks

Didn't know about the possibility to set variables.

mwilck:
LGTM

PR #68, opened 2022-11-08
Split kernel scriptlets into separate sub-package

lnussel:
Allows to replace them with noops for systems based on file triggers only

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -64,7 +64,9 @@ Requires:       rpm
 Requires(post): /usr/bin/grep
 Requires(post): /usr/bin/sed
 Requires(post): coreutils
+%if 0%{?suse_version} < 1550
 Provides:       suse-kernel-rpm-scriptlets = 0
+%endif
 Provides:       udev-extra-rules = 0.3.0
 Obsoletes:      udev-extra-rules < 0.3.0
 Provides:       system-tuning-common-SUSE = 0.3.0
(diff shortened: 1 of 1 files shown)


lnussel:
test build with kernel and prjconf modifications in https://build.opensuse.org/project/show/home:lnussel:branches:network
still takes time to finish
See also https://bugzilla.opensuse.org/show_bug.cgi?id=1205149

lnussel:
the failed obs build for 15.4 look unrelated to the change

mwilck:
Yes indeed, the 15.4 failure is unrelated.

So how is the prjconf change going?

lnussel:
release engineering will have to take care of that for the staging and when accepting the submission. Nothing we can do here other than telling them.

lnussel:
ping?

mwilck:
pong - I haven't followed the release engineering process, did you?

lnussel:
just submit then point the release engineer in charge at https://bugzilla.opensuse.org/show_bug.cgi?id=1205149#c2

mwilck:
I'd like to reverse the order of steps here. Last time my submission was rejected because of the build failures in question. And tbh I don't even know who the release engineers would be that I have to contact. Dominique?

lnussel:
for example, yes. Both Fabian and Dominique are CC'd in the bug so they are aware and didn't object the suggestion there.

mwilck:
Well, let's see how it goes, then.

PR #70, opened 2023-03-21
rpm-script: run scriptlets in /etc/kernel/{postinst,postrm}.d

mwilck:
This is useful e.g. for DKMS, see bsc#1194723.

We run scripts in {/usr,}/etc/kernel/postinst.d in %post, and {/usr,}/etc/kernel/postrm.d in %postun. The /usr directories are included to make it possible to install scripts from packages without touching /etc.

Note: /etc/kernel/xyz will _not_ override /usr/etc/kernel/xyz; both scripts will be run.

See https://kernel-team.pages.debian.net/kernel-handbook/ch-update-hooks.html for similar functionality on Debian.

### DIFF
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -104,6 +104,52 @@ run_bootloader () {
     fi
 }
 
+log() {
+    [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo "$*" >&2 || :
+}
+
+# Don't run arbitrary stuff:
+# - reject symlinks
+# - only run programs owned by root, and writable by noone but root
+check_perm() {
+    [ ! -L "$1" ] || return 1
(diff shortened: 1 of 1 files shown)


kernel-scriptlets/rpm-script
```diff
@@ -283,6 +331,8 @@ EOF
 	fi
 
 	[ -z "$certs" ] || /usr/lib/module-init-tools/kernel-scriptlets/cert-$op --ca-check 1 --certs "$certs" "$@"
+
+	run_parts /usr/etc/kernel/postinst.d /etc/kernel/postrm.d
```
hramrach, line 335: both postrm?

hramrach (changes requested):

hramrach:
Also doesn't DKMS need that uname_r passed to the script ?

mwilck:
Thanks for the review. Based on the recent discussion on BZ, I conclude this was wrong and short-sighted and I'll close it.

PR #71, opened 2023-06-02
Do not blacklist erofs

DaanDeMeyer:
erofs is actively maintained upstream and should not be considered a legacy filesystem so let's not blacklist it by default.

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -37,7 +37,7 @@
 %global systemd_units %{?with_boot_sysctl:boot-sysctl.service} %{?with_kernel_sysctl:kernel-sysctl.service}
 
 # List of legacy file systems to be blacklisted by default
-%global fs_blacklist adfs affs bfs befs cramfs efs erofs exofs freevxfs hfs hpfs jfs minix nilfs2 ntfs omfs qnx4 qnx6 sysv ufs
+%global fs_blacklist adfs affs bfs befs cramfs efs exofs freevxfs hfs hpfs jfs minix nilfs2 ntfs omfs qnx4 qnx6 sysv ufs
 
 # List of all files installed under modprobe.d
 # Note: this list contains files installed by previous versions, like 00-system-937216.conf!


hsiangkao:
Honestly, I have no idea why it was in the blacklist at all.

mwilck:
This list is the result of an internal discussion at SUSE. It basically means that SUSE doesn't support the file system in its proprietary offerings. openSUSE users are free to unblacklist these filesystems by simply `modprobe`'ing them.

@jeffmahoney, what do you think about erofs?

hsiangkao:
> This list is the result of an internal discussion at SUSE. It basically means that SUSE doesn't support the file system in its commercial offerings. openSUSE users are free to unblacklist these filesystems by simply `modprobe`'ing them.
> 
> @jeffmahoney, what do you think about erofs?

Ok, from my own perspective, I've been working on EROFS for more than five years and it already gains more users and developers.
It's still explicitly blocked from what I don't know (compared with other unmaintained stuffs on the list, honestly), maybe EROFS is still not quite good so I have to do more work to gain more and more users. Anyway, that is totally fine by me but a direction might be better for me to improve it for unblocking on SUSE.

mwilck:
@hsiangkao,

> maybe EROFS is still not quite good

Please don't interpret it this way. It's mostly a business decision, related to commercial support. It isn't directly related to the code quality, or what SUSE engineering perceives as such. If SUSE thought this module was "not good", we wouldn't be shipping it at all in the first place.

That said, the list is now about 4 years old, and it may be time to revisit it. Let's see what Jeff says.

jeffmahoney:
I haven't been following file system development for some time.  @goldwynr leads the file system team now.

goldwynr:
I dont' see a problem keeping EROFS for opensuse. It is acrively being developed and maintained upstream.

jeffmahoney:
With Leap inheriting binary packages from SLE, the changes apply to both.

goldwynr:
In that case, we would keep it blacklisted for both openSUSE and SLE. We don't officially support it. If someone wants to use it, the user can still explicitly modprobe it.

DaanDeMeyer:
Any chance that this could be a little more lenient in Tumbleweed? I'd be happy to have it there, I understand that it can't be supported in the enterprise offerings.

mwilck:
We can discuss different rules for Tumbleweed.

But as mentioned before, the idea is that the user needs to load the module explicitly once using "modprobe", indicating that she actually wants to use it (this way we avoid the module being auto-loaded without the user's consent, at least the first time). Then the following happens:

```
# modprobe erofs
unblacklist: loading erofs file system module
unblacklist: Do you want to un-blacklist erofs permanently (<y>es/<n>o/n<e>ver)? y
unblacklist: erofs un-blacklisted by creating /etc/modprobe.d/60-blacklist_fs-erofs.conf
# ls -l /etc/modprobe.d/60-blacklist_fs-erofs.conf
lrwxrwxrwx 1 root root 9 Jun  2 23:07 /etc/modprobe.d/60-blacklist_fs-erofs.conf -> /dev/null
```

Is that not acceptable?

bluca:
This is about automated usage at early boot (ie: root or /usr dm-verity partition), so afraid not

mwilck:
> This is about automated usage at early boot (ie: root or /usr dm-verity partition), so afraid not

As I said, you just need to excecute modprobe **once**. After that, automated usage will be possible.

DaanDeMeyer:
> As I said, you just need to excecute modprobe once. After that, automated usage will be possible.

This is specifically for use in CI testing where we throw away the image every time, manually running modprobe is not an option.

hsiangkao:
Hi, personally I think it'd be better to drop EROFS from the blacklist in openSUSE community version (and surely, it shouldn't have some relationship with SLE commercial support.)

After I digged into some, I found the original purpose of this change:
https://github.com/openSUSE/suse-module-tools/pull/5/commits/8cb42fb6658f210cb8c955d584a65f7b041c0575

but EROFS actually fixes all syzbot upstream issues in time and the kernel support actually does good on malicious images, see:
https://syzkaller.appspot.com/upstream/s/erofs

compared with something like jfs or reiserfs:
https://syzkaller.appspot.com/upstream/s/jfs
https://syzkaller.appspot.com/upstream/s/reiserfs

also compared with
https://syzkaller.appspot.com/upstream/s/ntfs3

Child subsystems: 9p (6) afs (1) autofs (1) bfs (1) btrfs (51) cachefs (0) erofs (0) ext4 (47) f2fs (16) fat (8) fscrypt (1) fsverity (0) fuse (6) gfs2 (18) hfs (46) isofs (0) jfs (59) kernfs (8) nfs (0) nilfs (13) ntfs (24) ntfs3 (57) overlayfs (8) reiserfs (73) squashfs (4) udf (14) xfs (22)

Yes, EROFS is still actively developping, so there might be some new syzbot fuzzed image occurence, but we always fix these reports in time.

DaanDeMeyer:
> ln -s /dev/null /etc/modprobe.d/60-blacklist_fs-erofs.conf

Our test images only ship /usr, so that doesn't work either. Currently we just remove the file, which works. This issue is about gradually getting rid of as many customizations to the test image as possible by fixing things upstream instead of working around issues locally.

mwilck:
> Hi, personally I think it'd be better to drop EROFS from the blacklist in openSUSE community version 

I hear you, and I have no doubt that the upstream maintainers doing an excellent job fixing bugs in erofs. Again: This is not about the job the erofs maintainers are doing, or the quality of the code.

I've tried to explain that for real-world users, the issue is fixed with one single shell command. I agree that it could be documented better, but other than that, disabling the blacklist entry is a pretty straightforward thing.

Please think of this list more like an "inverted set of supported file systems". SUSE officially supports a certain, relatively small set of file systems. Usually this means that SUSE deems these file systems important for its business, and thus has in-house expertise for these file systems in one way or the other. 

erofs is currently not in this "whitelist". The set of supported file systems is not the decision of the `suse-module-tools` maintainer, therefor a PR like this is not the right place to make changes to the list happen.

mwilck:
> Our test images only ship /usr, so that doesn't work either.

Use `/run`, then?

> This issue is about gradually getting rid of as many customizations to the test image as possible

This is understandable, but our options are limited. What exactly are you testing?

hsiangkao:
> Please think of this list more like an "inverted set of supported file systems". SUSE officially supports a certain, relatively small set of file systems. Usually this means that SUSE deems these file systems important for its business, and thus has in-house expertise for these file systems in one way or the other.

Nope, you misparse my question. May I ask if OpenSUSE is a community version? Why does OpenSUSE scare a community user when they try to use EROFS and make them run away? I don't find fedora or debian behaves like this.

DaanDeMeyer:
It would also be nice if the comment could be updated that this is not about legacy filesystems but about supported filesystems

bluca:
If the issue is commercial support in SLES, wouldn't it be better if the blocklist was shipped in a SLES-specific package, so that openSUSE users do not have to do manual workarounds?

mwilck:
The modified wording is now in the master branch.

Meanwhile I initiated an internal discussion about the future handling of these blacklist entries, in particular for openSUSE.
Let's see how it plays out.

PR #72, opened 2023-06-05
rpm-script: skip run_bootloader check (boo#1208117)

mwilck:
It would be better if `run_bootloader()` didn't make any assumption on the supported bootloaders and let /usr/lib/bootloader/bootloader_entry figure out itself what to do with the current installation. This way other boot loaders such as systemd-boot can be more easily supported in the future.


hramrach (approved):
Looks good

PR #73, opened 2023-07-03
rpm-script: check for regenerate-initrd-posttrans in %posttrans

mwilck:
rpm-script: check for regenerate-initrd-posttrans in %posttrans (boo#1212957)

If zypper is run with INITRD_IN_POSTTRANS=1 set, no initrd will be
generated. Setting this variable globally is useful in some situations,
in particular where split kernel packages like kernel-defaukt-extra are
in use. Checking the regenerate-initrd-posttrans flag is cheap if the
flag isn't present, and will fix the issue if it is.


hramrach (approved):
LGTM

Thanks

PR #75, opened 2023-07-25
rpm-script: detect usrmerged automatically

lnussel:
Don't rely on the kernel rpm telling us it's usrmerge, just look at the facts instead. Related to boo#1206798

### DIFF
--- a/kernel-scriptlets/inkmp-script
+++ b/kernel-scriptlets/inkmp-script
@@ -8,7 +8,6 @@ release=""
 kernelrelease=""
 flavor=""
 variant=""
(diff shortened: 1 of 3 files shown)


lnussel:
Do you actually use the same code on SLE also? If you are using a branch we could also simply get rid of the non usrmerge path

mwilck:
I am using branches. The `kernel-scriptlets` used to be identical, but they have diverged between the SLE branches and the master branch.

So yeah, we can just get rid of the `[ -L /lib ]` test. I'm going to merge this and apply the change on top.

PR #76, opened 2023-07-31
rpm-script: update bootloader after creating initramfs

mwilck:
boo#1213822: updating the bootloader may create invalid entries
for kernels which don't have an initramfs (yet). For example,
grub2's "10_linux" script doesn't use root=UUID= boot parameter
for kernels without initramfs, and uses ${GRUB_DEVICE} instead,
which may result in an invalid entry e.h. for multi-device btrfs
partitions, like this:

        echo    'Loading Linux 6.4.6-1-default ...'
        linux   /boot/vmlinuz-6.4.6-1-default root=/dev/sda7
/dev/sda5  ${extra_cmdline} crashkernel=128M@256M showopts

Fix it by updating the bootloader in posttrans if
INITRD_IN_POSTTRANS=1 is set.

Fixes: 4fadbee ("rpm-script: check for regenerate-initrd-posttrans in %posttrans (boo#1212957)")

### DIFF
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -90,6 +90,37 @@ message_install_bl () {
     echo "available bootloader for your platform (e.g. grub, lilo, zipl, ...)."
 }
 
+update_bootloader() {
+    if [ ! -e /.buildenv ] ; then
+	if [ -f /etc/fstab ] ; then
+	    # only run the bootloader if the usual bootloader configuration
+	    # files are there -- this is different on every architecture
+	    initrd=initrd-"$kernelrelease"-"$flavor"
+	    if [ "$flavor" = rt ]; then
+		default=force-default
+	    fi
+	    # Note: the 2nd condition is for removing the bootloader
+	    # entry for an uninstalled kernel.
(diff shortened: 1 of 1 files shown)


hramrach (approved):
LGTM

This will fail to remove ramdiskless kernels from the bootloader menu (will be re-added in posttrans if KMPs built against the kernel exist) which was not supported in the first place - the remove part in postun always specifies a ramdisk.

One more reason to move KMPs out of the kernel module directory.

mwilck:
Thanks.

Do we support ramdisk-less kernels, or do we plan to support them in the future?

Wrt KMPs, wm2 doesn't blindly rebuild initrds for any KMP. It does only if the KMP uses `%kernel_module_package -b`, _or if no initrd for the given kernel exists_. The latter condition must obviously be removed if we want to support ramdisk-less kernels, but that's independent of this PR.

hramrach:
I don't think ramdiskless kernels work. The upstream tools support them but they are not tested in the distribution so there are likely some other cases where they are not handled correctly.

hramrach:
Currently KMP modules are installed into the module directory of the kernel against which they build (even if they can be used with different kernel versions) and the scripts test existence of this directory.

mwilck:
Merged as 8344155

PR #77, opened 2023-08-14
splitoff regenerate-initrd-posttrans

mwilck:
- kernel-scriptlets: assume usrmerged distribution
- split off regenerate-initrd-posttrans in a separate package

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -224,7 +224,6 @@ exit 0
 %dir /usr/lib/module-init-tools
 /usr/lib/module-init-tools/driver-check.sh
 /usr/lib/module-init-tools/lsinitrd-quick
-/usr/lib/module-init-tools/regenerate-initrd-posttrans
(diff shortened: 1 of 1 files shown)


lnussel:
regenerate-initrd-posttrans wouldn't be pulled in by any package though?

mwilck:
I thought that would be achieved by the split-alias `Provides:` tag. Did I get that wrong?

lnussel:
that works on `zypper dup` as long as recommends are turned on. It would not do anything on new installations.

mwilck:
Thanks. I wasn't aware that this works only with recommends. I'll re-work it.

mwilck:
Re-pushed, we have the simplest possible approach now. But we don't handle the case where the initramfs is built directly in `%post`.

lnussel (approved):
lgtm

mwilck:
merged in 16.0.35

PR #78, opened 2023-08-14
Revert "rpm-script: detect usrmerged automatically"

hramrach:
This reverts commit 3a6e48605b1dc4637d51c71ca788ed4fd46e5c42.

The [ -L /lib ] sounds less reliable than the kernel to telling the scripts when the modules are placed in /usr.

The kernel always knows where it placed the modules.

On the other hand, the /lib symlink only exists when a specific package is installed. This may not be the case due to package installation ordering or in special environments not directly managed by rpm.

### DIFF
--- a/kernel-scriptlets/inkmp-script
+++ b/kernel-scriptlets/inkmp-script
@@ -8,6 +8,7 @@ release=""
 kernelrelease=""
 flavor=""
 variant=""
+usrmerged="0"
 image=""
 certs=""
 
@@ -40,7 +41,7 @@ while true ; do
 	    ;;
 
 	--usrmerged)
-	    # ignored for legacy compat reasons
+	    usrmerged="$2"
 	    shift
 	    ;;
 	--image)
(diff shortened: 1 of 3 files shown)


lnussel:
Agreed with Martin that the code should just not do any tests and assume usrmerge in Factory always.

hramrach:
As in the script should be hardcoded at s-m-t build time to do one or the other?

lnussel:
yes, even though there is no "other" in the Factory ie master branch. Factory is usrmerged. There is no expectation to be able to use the Factory package on SLE.

hramrach:
The code diverging between Factory and SLE will be annoying for backporting fixes/updates, and probably error-prone, too.

lnussel:
It's two schools of thought I guess. Either you keep compat code around and some runtime/buildtime detection or use a branch and apply potential fixes where needed. Martin seems to prefer a branch, cherry picking fixes. Which is also what I think makes most sense here. For the kernel you've decided to maintain package files in a different way which also comes with a set of problems like risking to break stable code stream with supposedly Factory only changes. btdt. choose your poison.

PR #79, opened 2023-08-15
usrmerge: Drop support for non-usrmerged.

hramrach:
The master branch is only for usrmerged system, drop compatibility.

Link: https://github.com/openSUSE/suse-module-tools/pull/78


hramrach:
See also the `non-usrmerged` branch for comparison

lnussel (approved):

suse-module-tools.spec
```diff
@@ -79,17 +67,14 @@ Recommends:     kmod
 # This release requires the dracut module 90nvdimm
 Conflicts:      dracut < 49.1
 # TW: conflict with pre-usrmerge
-%if 0%{?suse_version} >= 1550
-Conflicts:      filesystem < 15.5-40.2
-%endif
+Conflicts:      filesystem < 16
```
mwilck, line 70: Where did you get this from?

hramrach: The version in Leap 15.5 is 15.0, in TW is 84, ALP is identical.

Rather I would ask where that odd number is from, and the answer is likely some confusion during early usrmerge implementation.

hramrach:
Pushed the corresponding branch without usrmerge branch as sle1s5-sp6 - can't make it a PR so feel free to massage it afterwards.

hramrach:
What is the fallout, exactly?

Does the Base:System project not warn not to be used for SLE?

mwilck:
The problem is that we have the factory version of s-m-t in Base:System, which conflicts with the filesystem package in SLE.

mwilck:
```
15.5                 aarch64    unresolvable: 
      conflict for providers of filesystem needed by glibc
      (provider filesystem is in conflict with suse-module-tools)
```

hramrach:
That's correct: the packages in Base:System should not be used in SLE, and if they conflict with SLE the better.

hramrach:
> This is the Base:System devel project. Its main purpose is to serve as development project for packages around the topic of the system base in the openSUSE:Factory distribution. If you want to participate you can contact us individually or use our mailinglist opensuse-factory@opensuse.org. Please report bugs to the respective bugowners as set.
> 
> Warning: This is not a backports project! Installing the packages here on an old distribution may cause severe breakage!
> 
> DO NOT ENABLE BUILDING FOR OLD DISTROS!

mwilck:
I know. Still, `Base:System` is a test bed for new package versions and having it build packages for SLE is convenient. I know this can be handled by just linking the package to some backport project, but that's pretty inconvenient. And of course, the failure messages cast doubt on the package quality.

Anyway, currently just noting this. I am not sure if it warrants reverting your patch.

hramrach:
You can set s-m-t to not build for SLE in Base:System. Then you will not get any error.

It is normal that some Tumbleweed packages do not work on SLE, that does not warrant a revert. After all this repo has a separate branch for each release on the basis that compatibility is not required.

I proposed compatible changes and they were rejected for this very reason.

PR #82, opened 2023-08-21
wm2: Update for usrmerge

hramrach:
wm2 has many paths hardcoded to /lib/modules

Change to /usr/lib/modules

### DIFF
--- a/weak-modules2
+++ b/weak-modules2
@@ -2,11 +2,11 @@
 
 ##############################################################################
 # How it works:
-# * Kernels install modules below /lib/modules/$krel/kernel/.
-# * KMPs install modules below /lib/modules/$krel/updates/ or .../extra/.
+# * Kernels install modules below /usr/lib/modules/$krel/kernel/.
+# * KMPs install modules below /usr/lib/modules/$krel/updates/ or .../extra/.
 # * Symbolic links to modules of compatible KMPs are created under
-#   /lib/modules/$krel/weak-updates/{updates,extra}/... (the original path
-#   below /lib/modules/$other_krel is used).
+#   /usr/lib/modules/$krel/weak-updates/{updates,extra}/... (the original path
+#   below /usr/lib/modules/$other_krel is used).
 # * Depmod searches the directories in this order: updates/, extra/,
 #   weak-updates/, kernel/ (see /etd/depmod.conf or
 #   /etc/depmod.d/00-system.conf for details).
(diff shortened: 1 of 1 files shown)


hramrach:
Can't really test this on Tumbleweed.

We don't have the kernel ABI there, even if we had the KMPs are always built against the current kernel so wm2 does nothing there either way.

mwilck:
I'm not sure what this is going to buy us. For all I know, the `/lib → /usr/lib` symlink is going to stay for good. Is it just for cleaning up?

mwilck:
There seems to be a typo in the bugzilla reference?

hramrach:
Both kmod and rpm now refer to the modules by the /usr/lib path.

With wm2 referring to modules by /lib path it will report problem modules with the wrong path.

It also has to create the symlink in any temporary directories it creates.

That bug reference was actually correct, wm2 reported an error when crating the ramdisk which turned out to be irrelevant for the bug.

Nonetheless, it shows that it's not going to work as-is.

With this the error is gone but the functionality is not verified.

mwilck:
OK, I'll do another review. Thanks.

mwilck:
I sent you a PR at https://github.com/hramrach/suse-module-tools/pull/1.

I'll push a new s-m-t version without this, and see if we can find a way to test this code somehow next week or so.

hramrach:
I suppose the test would be to build some KMPs with modules installed somewhere other than the kernel module directory, and installing them.

mwilck:
Pending a test, I think this PR is good to go now. However I just pushed 16.0.34 (without this patch) to Factory.

hramrach:
A test package https://build.opensuse.org/package/show/home:michals/hdjmod (works correctly on Leap 15.5)

mwilck:
Something's weird with the TW build of the hdjmod package
```
Resolving package dependencies...

Problem: nothing provides 'kernel-uname-r = hdjmod-1.34-3.1' needed by the to be installed hdjmod-kmp-default-1.34_k6.4.12_1-3.1.x86_64
```

hramrach:
Maybe the rpm scripts are guessing the kernel release from the module filename. Then abolishing TW special handling in rpm-config-SUSE should fix it.

mwilck:
Looks like a wm2 issue
```
(1/1) Installing: hdjmod-kmp-default-1.34_k6.4.12_1-3.1.x86_64 ..................................................................[done]
depmod: ERROR: could not open directory /var/tmp/weak-modules2.o52Qsg/6.4.11-1-default/usr/lib/modules/6.4.11-1-default: No such file or directory
depmod: FATAL: could not search modules: No such file or directory
```

hramrach:
That's expected without this patch

mwilck:
okay, I'm typing more quickly than I can think... give me some time

hramrach:
Here .. but I would have to rebuild both the kernel and the KMP with patches r-c-S https://github.com/openSUSE/rpm-config-SUSE/pull/70

mwilck:
I tested this on TW by removing and re-adding `kernel-default-6.4.11-1.1.x86_64.rpm` multiple times. It worked correctly. So at least the "compatible KMP" case works.

mwilck:
LGTM, I'm going to merge this.

PR #85, opened 2023-08-21
Revert "weak-modules2: accept modules under /usr/lib/modules on stdin"

hramrach:
Modules under /usr not supported on SLE


hramrach:
Well, it does add annoying code obfuscation of which we got rid in Factory.

Getting rid of it on SLE as well is motivation for the revert.

PR #86, opened 2023-09-01
Sles15 sp6 bsc1214877

hramrach:
Probably needed on master as well.

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -34,7 +34,7 @@
 %global modprobe_conf_rpmsave %(echo "%{modprobe_conf_files}" | sed 's,\\([^ ]*\\),%{_sysconfdir}/modprobe.d/\\1.conf.rpmsave,g')
 
 Name:           suse-module-tools
-Version:        15.6.0
+Version:        15.6.1
 Release:        0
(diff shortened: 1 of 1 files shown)


mwilck:
I suppose it can't hurt, but I'd like to understand how the broken file name in `/run/regenerate-initrd came to exist in the first place.

hramrach:
Stripping the name would work only for broken value but would break correct value.

The image is something like vmlinuz-5.14.21-150400.24.81-default, and kver something like 5.14.21-150400.24.81-default.

If the full image value is to be supported as well as version only then checking the filename as is and then stripping it and checking it again is required.

mwilck:
> Stripping the name would work only for broken value but would break correct value.

Right, I just figured that, too. I'll add the comment change to the master branch for now. Wrt the actual issue, I tend to revert 60a2a14, which was a fix for f4187b6, which we don't have in SLE15, anyway.

hramrach:
As far as I can tell the API change is not required to fix the bug.

mwilck:
What API change do you mean? The actual API change, which likely caused bug 1214877, was 60a2a14, which I'm now proposing to revert.

hramrach:
Yes, reverting that for SLE sounds like a good thing.

Still even for TW/ALP the change is not great - see #89

PR #89, opened 2023-09-01
Fix kernel test in regenerate-initrd-posttrans (bsc#1214877).

hramrach:
commit 60a2a14f978e ("weak-modules2: only use kernel version under /run/regenerate-initrd") changes the API of regenerate-initrd-posttrans to only take the kernel version, not the image name.

It also checks that the module directory exists, not the kernel image.

This is both broken.

While the image is removed with the kernel the module directory may contain KMPs built against this kernel version and remain.

Anyway, at least revert the API change, there seem to be some users outside of suse-module-tools.

The problem with the remaining kernel module directory should be resolved by moving out-of-tree KMPs outside of the kernel module directory.

Fixes: 60a2a14f978e ("weak-modules2: only use kernel version under /run/regenerate-initrd")

### DIFF
--- a/regenerate-initrd-posttrans
+++ b/regenerate-initrd-posttrans
@@ -2,9 +2,9 @@
 #
 # Packages that install kernels or kernel-modules create a flag
 #
-#   /run/regenerate-initrd/<kernel version>
+#   /run/regenerate-initrd/<kernel image>
 #
-# to have the initrd for <kernel version> generated, or
+# to have the initrd for <kernel image> generated, or
 #
 #   /run/regenerate-initrd/all
 #
@@ -48,7 +48,8 @@ for f in "$dir"/*; do
 		[ -e "$f" ] || break;;
 	esac
 	rm -f "$f"
-	kver=${f##*/}
+	image=${f##*/}
+	kver=${image#*-}
 	[ -d /lib/modules/"$kver" ] || {
 	    echo $0: skipping invalid kernel version "$dir/$kver"
 	    continue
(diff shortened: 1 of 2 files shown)


mwilck:
We can't do this, it leads to known breakage (bsc#1201387). We can solve the breakage differently, but I am pretty certain that using the kernel version as marker here is actually the right thing to do. 

It says "rebuild the initrd for version $X", which is what we want, and it does so without depending on any knowledge about the various image names and installation paths for different architectures and OS versions. Note that we simply call `dracut --kver $X` and afaik dracut also checks for the existence of the module directory, not for the kernel image.

We can discuss about removing the check of the directory existence completely. I think it only exists to avoid errors from dracut.

I'm not sure what happens if the module directory exists but the kernel image doesn't. dracut will try to build the intird. If not only the image is missing but also the kernel's modules, it will likely fail. But if only the image has been deleted or renamed, it will just work.

Also note that rebuilding the initrd for a kernel that has been removed is very likely a mistake by itself.

Wrt other users of this "API", we've had no issues with this in factory for more than a year. If we find out that kernel live patching has been using this "API" without me knowing about it, we can fix it in time for ALP.

hramrach:
Why can't?

This only changes the external API, back to what it was, it does not change that the module directory is checked, not the image.

With checking the module directory rather than the image there is no 'knowledge' of the image name other than it does not contain a hyphen.

mwilck:
I have pushed my attempt to fix this as #90.

PR #90, opened 2023-09-01
support old-style markers under /run/regenerate-initrd

mwilck:
With a235a6e ("weak-modules2: only use kernel version under /run/regenerate-initrd")
we changed the API for programs for creating markers under
/run/regenerate-initrd. While the main user of this API is weak-modules2,
there may be external tools relying on the old behavior. Make sure
the old style, where the name of the kernel image was as marker.

Fixes: bsc#1214877

### DIFF
--- a/regenerate-initrd-posttrans
+++ b/regenerate-initrd-posttrans
@@ -49,6 +49,11 @@ for f in "$dir"/*; do
 	esac
 	rm -f "$f"
 	kver=${f##*/}
+	case "$kver" in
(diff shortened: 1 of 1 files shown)


hramrach:
You were against knowledge of image names yet here it is.

Can't there be test of the name as is, and then with the part up to the first hyphen stripped without examining the exact format?

hramrach:
hm, it could be checked without examining that the prefix looks like kernel image name but it does not provide a way to construct the error message.

hramrach (approved):
LGTM

Thanks

PR #91, opened 2023-09-22
Use package dependencies, not binary paths (bsc#1215533)

hramrach:
OBS cannot automatically resolve binary paths when installeing packages leading to errors when package is instlled in OBS

### DIFF
--- a/suse-module-tools.spec
+++ b/suse-module-tools.spec
@@ -19,8 +19,8 @@
 Name:           suse-module-tools
 Version:        12.11
 Release:        0
-Requires:       /sbin/depmod
-Requires:       /sbin/mkinitrd
+Requires:       kmod-compat
+Requires:       dracut
 Requires:       findutils
 Requires:       gzip
 # for grepping /etc/SUSE-release
@@ -29,7 +29,7 @@ PreReq:         grep
 # in post, it also requires getopt (coreutils) and sed
 PreReq:         coreutils rpm
 # XXX: this should be nm OR eu-nm, the script works with both
-PreReq:         /usr/bin/eu-nm /bin/sed
+PreReq:         elfutils sed
 Summary:        Configuration for module loading and SUSE-specific utilities for KMPs
 License:        GPL-2.0-or-later
 Group:          System/Base


mwilck:
These dependencies were added for good reasons (although I currently don't remember the details, and have no time to dig them up, sorry).

If you have issues in particular OBS projects, use `FileProvides: ` statements in the prjconf. This is how I'm told these deficiencies of OBS should be handled.

See https://openbuildservice.org/help/manuals/obs-user-guide/cha.obs.prjconfig.html (search for "FileProvides").

mwilck:
**s-m-t** has a long history of replacing package dependencies by file requires and vice versa. I tend to keep my hands off them, because I don't fully understand the rationale. I agree that in general, package requires are cleaner and cause less headaches in OBS. But various people who know more about the distribution architecture than myself have suggested using file requires in the past. This includes @thkukuk and (IIRC) @DimStar77.

Arguments for file requires that I remember:
* sometimes needed if a spec file must work on different distros and package names have changed (this is not an issue here, obviously, as we're talking about a SLE12 specific branch)
* makes it possible to reduce system size for small images (see 3b34126 ("spec: rework dependencies"))

There may be more.

Given that this is SLE12-SP5, I am reluctant to changes in this area, even though I understand that the problem you reported will likely affect any branch project. Anyway this can be worked around on IBS using `prjconf`, see above.

I'll keep this open for a while in case some of the experts want to comment and enlighten me more about it. At least for the factory branch I'd like to have rationales for every file requires in the spec file in the form of comments, lest I forget again what they are good for.

thkukuk:
Package requires are good, if you need one specific package.
File requires are the way to go, if you need a file (config file, data file, binary), which is provided by several packages, but you don't care which implementation is used. E.g. if sed from `sed` package or `busybox` is used.
Or if you need /usr/bin/uptime and you don't care which implementation.

In this case, I think mkinitrd switched several times the package, so a file requires is easier.
eu-nm should only exists in elfutil, so a require for the package is ok.

mwilck:
Closing this per @thkukuk's comment. The workaround is to use Fileprovides in prjconf in OBS (see above).

PR #92, opened 2023-09-25
modprobe.d: Blacklist cls_tcindex module (bsc#1210335)

Werkov:
This should go into SLE15-SP5, SLE15-SP4 and SLE15-SP3-LTSS, SLE15-SP2-LTSS.
Not sure what is the way to submit here, so starting with the first one.


mwilck:
I'm going to merge this in all branches.

mwilck:
Merged.

PR #93, opened 2023-09-29
Drop support for INITRD_MODULES (jsc#PED-1915)

mwilck:
- weak-modules2: ignore INITRD_MODULES (jsc#PED-1915)
- weak-modules2: obtain list of initrd modules from dracut (jsc#PED-1915)

### DIFF
--- a/weak-modules2
+++ b/weak-modules2
@@ -390,10 +390,70 @@ previous_version_of_kmp() {
     return $res
 }
 
-get_initrd_basenames() {
-    $LSINITRD /boot/initrd-$1 | \
-	sed -rn 's:.*\<usr/lib/modules/.*/::p' | \
+# Create list of dracut configuration files to read, taking into
+# account priorities of the user and built-in configuration.
+# Copied from dracut.sh as of dracut 059 (GPL-2.0-or-later)
+dropindirs_sort() {
+    local suffix=$1
+    shift
+    local -a files
+    local f d
+
+    for d in "$@"; do
+        for i in "$d/"*"$suffix"; do
+            if [[ -e $i ]]; then
+                printf "%s\n" "${i##*/}"
+            fi
+        done
+    done | sort -Vu | {
+        readarray -t files
+
+        for f in "${files[@]}"; do
+            for d in "$@"; do
+                if [[ -e "$d/$f" ]]; then
+                    printf "%s\n" "$d/$f"
+                    continue 2
+                fi
+            done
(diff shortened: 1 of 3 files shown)


mwilck:
@aafeijoo-suse for information

weak-modules2
```diff
@@ -390,12 +390,72 @@ previous_version_of_kmp() {
...
+                    continue 2
+                fi
+            done
+        done
+    }
+}
+
+get_current_basenames() {
+    $LSINITRD /boot/initrd-$1 |
+	sed -rn 's:.*\<usr/lib/modules/.*/::p' |
```
aafeijoo-suse, line 424: I think you accidentally removed two `\`, at the end of lines 423 and 424

weak-modules2
```diff
@@ -390,12 +390,72 @@ previous_version_of_kmp() {
...
+
+    # Assemble the content of the dracut configuration files.
+    # Make sure to put a newline between subsequent conf file contents.
+    for cf in $(dropindirs_sort .conf "$DRACUT_CONFDIR" "$DRACUT_BUILTIN_CONFDIR"); do
+	conf="$conf
+$(cat "$cf")"
+    done
+
+    # run get_dracut_drivers with reduced privileges
+    get_current_basenames "$1" |
```
aafeijoo-suse, line 454: Is this pipe ok?

weak-modules2
aafeijoo-suse: Oh, I see it's not necessary. Sorry for the noise.

weak-modules2
mwilck: Why not?

aafeijoo-suse: hmm... ok, the issue is in `get_dracut_drivers`, if there is no `omit_drivers` in any conf file, it does not print any modules.

mwilck: Wow, good catch. Thanks.

mwilck: Should be fixed with the fixup I just pushed.

aafeijoo-suse: You're welcome. Yes, now it looks good.

mwilck:
Rebased and squashed the fixup patch.

mwilck:
@hramrach says he feels not familiar enough with the subject to review. So I guess we just move forward. Hardly anyone will be using `INITRD_MODULES` in factory anyway, I suppose.

PR #94, opened 2023-10-02
modprobe.d: Blacklist cls_tcindex module (bsc#1210335)

Werkov:
This is addition to #92 since 4.12-based kernels are affected too.


mwilck:
merged in 12.13

PR #99, opened 2024-01-11
macros.initrd: %regenerate_initrd_post: don't fail if mkdir is unavailable

mwilck:
mkdir is part of the coreutils package. When the post scriptlet of some
package using regenerate_initrd_post is run, the mkdir command may be
unavailable. This is non-fatal, because in all cases that matter for initrd
generation, coreutils will be part of the same transaction and will trigger
an initrd rebuild later. See boo#1217775.

Anyway, packages can use the regenerate_initrd_requires macro below to make
sure mkdir is available in their post scriptlet. Use if this macro is not
mandatory.

### DIFF
--- a/macros.initrd
+++ b/macros.initrd
@@ -3,9 +3,23 @@
 # regenerated
 # See also fate#313506
 
+# mkdir is part of the coreutils package. When the post scriptlet of some
+# package using regenerate_initrd_post is run, the mkdir command may be
+# unavailable. This is non-fatal, because in all cases that matter for initrd
+# generation, coreutils will be part of the same transaction and will trigger
+# an initrd rebuild later. See boo#1217775.
+# 
+# Anyway, packages can use the regenerate_initrd_requires macro below to make
+# sure mkdir is available in their post scriptlet. Use if this macro is not
+# mandatory.
+
+%regenerate_initrd_requires \
+Requires(post): /usr/bin/mkdir \
+%nil
+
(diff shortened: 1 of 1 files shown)


mwilck:
See also https://github.com/openSUSE/rpm-config-SUSE/pull/71

macros.initrd
```diff
@@ -3,9 +3,23 @@
...
+# mandatory.
+
+%regenerate_initrd_requires \
+Requires(post): /usr/bin/mkdir \
+%nil
+
 %regenerate_initrd_post \
-	mkdir -p /run/regenerate-initrd/ \
-	touch /run/regenerate-initrd/all \
+        ! command -v mkdir >/dev/null || mkdir -p /run/regenerate-initrd/; \
```
thkukuk, line 21: In other packages we had big problems with "command -v" and had to use "test -x /usr/bin/mkdir" instead.
But I'm neither a shell expert nor did I understand the reason for the problem, so I cannot say that this is safe or not.

mwilck: Hm, "command" is part of POSIX and a regular built-in in both bash and dash. 
I'd be grateful if you could point me to one of the issues you saw.

mwilck: Quote from @thkukuk on bugzilla: "Ok, I understand now the problem with the other packages, the above PRs are fine for me and should not have that problem."

PR #100, opened 2024-03-11
weak-modules2: update static nodes after running depmod

mwilck:
When modules are installed or removed for the currently running
kernel, recreate static nodes to make enable the functionality
(bsc#1220303).


mwilck:
This would possibly overwrite existing static device nodes. We can't implement this feature like this.

PR #102, opened 2024-05-14
regenerate-initrd-posttrans: run update-bootloader --refresh for XEN

mwilck:
On XEN configurations, the newly created initrds must be copied to
the EFI system partition. update-bootloader --refresh (actually,
grub2-mkconfig via /etc/grub.d/20_linux_xen) will take care of this.
See bsc#1223278.

### DIFF
--- a/regenerate-initrd-posttrans
+++ b/regenerate-initrd-posttrans
@@ -1,4 +1,8 @@
 #!/bin/sh
+# Local variables:
+# indent-tabs-mode: t
+# sh-basic-offset: 8
+# End:
 #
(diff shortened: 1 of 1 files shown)


WenhuaChang:
LGTM too.

mwilck:
> It looks generally OK but the use of `||` in place of `if` is very hard to read.

It's a matter of taste. Personally I like this style and find it actually quite easy to understand. Also, it matches the current style of `regenerate-initrd-posttrans`. So, unless there are serious objections, I'd like to keep it this way for now.

PR #105, opened 2024-07-03
shellcheck fixes

mwilck:
With this PR, the kernel scriptlets pass shellcheck

It contains also a few other fixes and minor code reordering.
No change of functionality.

- **rpm-script: fix minor issues reported by shellcheck**
- **rpm-script: move arm64 page-size check to separate function**
- **rpm_script: further simplify check_arm_pagesize()**
- **rpm-script: move cert-script invocation to a shell function**
- **inkmp-script: fix warnings reported by shellcheck**
- **kmp-script: fix warnings reported by shellcheck**
- **cert-script: annotate some shellcheck warnings**

### DIFF
--- a/kernel-scriptlets/inkmp-script
+++ b/kernel-scriptlets/inkmp-script
@@ -61,9 +61,6 @@ done
 wm2=/usr/lib/module-init-tools/weak-modules2
 nvr="$name"-"$version"-"$release"
 
-modules_dir=/usr/lib/modules/$kernelrelease-$flavor
-system_map=${modules_dir}/System.map
-
 run_wm2() {
     [ -z "$KERNEL_PACKAGE_SCRIPT_DEBUG" ] || echo wm2 "$@" >&2
     $wm2 "$@"
@@ -81,7 +78,7 @@ case $op in
 	;;
     post)
 	if [ -x "$wm2" ]; then
-	    rpm -ql "$nvr" | INITRD_IN_POSTTRANS=1 run_wm2 --add-kernel-modules $kernelrelease-$flavor || script_rc=$?
+	    rpm -ql "$nvr" | INITRD_IN_POSTTRANS=1 run_wm2 --add-kernel-modules "$kernelrelease-$flavor" || script_rc=$?
 	fi
 	;;
     preun)
@@ -98,7 +95,7 @@ case $op in
 	# This is similar to the check for $system_map in rpm-script.
 	for __i in "${!modules[@]}"; do
 	    if [[ -e "${modules[$__i]}" ]]; then
-		unset -v modules["$__i"]
+		unset -v 'modules["$__i"]'
 	    fi
 	done
 	if [[ "${#modules[@]}" == 0 ]]; then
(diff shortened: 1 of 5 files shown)


aplanas (approved):
LGTM

aplanas (approved):

suse-module-tools.spec
```diff
@@ -60,8 +60,6 @@ Provides:       udev-extra-rules = 0.3.0
 Obsoletes:      udev-extra-rules < 0.3.0
 Provides:       system-tuning-common-SUSE = 0.3.0
 Obsoletes:      system-tuning-common-SUSE < 0.3.0
-Provides:       sdbootutil-rpm-scriptlets = 2.0
-Obsoletes:      sdbootutil-rpm-scriptlets < 2.0
```
aplanas: Why is this removed?

mwilck: It was wrong. The dependencies must be set for the scriptlets package, not for the main package. I fixed this in "master" already.

mwilck:
I did some basic testing with this and it seems to work ok.

@aplanas any tests with real FDE / sdbootutil systems would be very welcome.

suse-module-tools.spec
mwilck: ah ... and I must have made some mistake rebasing this onto master, after I'd fixed the dependencies in the master branch.

kernel-scriptlets/inkmp-script
```diff
@@ -98,15 +95,15 @@ case $op in
 	# This is similar to the check for $system_map in rpm-script.
 	for __i in "${!modules[@]}"; do
 	    if [[ -e "${modules[$__i]}" ]]; then
-		unset -v modules["$__i"]
+		unset -v 'modules["$__i"]'
```
hramrach, line 98: This looks like a change in semantics

aplanas: That confused me too, but I checked it. Seems that it a removal of the element in the hash

mwilck: I was wondering about it, too, but I tested it and found it's indeed correct.

See https://www.shellcheck.net/wiki/SC2184 

I've seen lots of weird things in bash, but this is surely on the top ten list.

mwilck: reproducer:
```
> foo1=set
> echo $foo1
set
> foo=(a b c d)
> echo "${#foo[@]}"
4
> x=1
> touch foo1
> unset foo[$x]
> echo $foo1

> echo "${#foo[@]}"
4
> foo1=set
> unset 'foo[$x]'
> echo $foo1
set
> echo "${#foo[@]}"
3
```

mwilck: Relevant part in bash man page: "*When using a variable name with a subscript as an argument to a command, such as with unset, without using the word expansion syntax described above,  the  argument is subject to pathname expansion.  If pathname expansion is not desired, the argument should be quoted.*"

PR #106, opened 2024-07-15
Add support for mkosi-initrd

aafeijoo-suse:
`mkosi` provides the `mkosi-initrd` script [1], which allows to generate initrds locally on the system. Although it is not equivalent in features to `dracut`, it's a valid alternative with its own benefits in some use cases:
- Clear ownership of bugs: no 3rd party binary (e.g. dracut) deciding what is installed from each package or even modifying them.
- Any change on a package configured to be included would automatically apply to the initrd (if the package triggers initrd rebuild).
- It attempts to shift the maintenance effort from the initrd generator to the responsible package.

Currently, `kernel-install` [2] allows to specify the initrd generator to use via `initrd_generator=` in `install.conf`, so make use of this configuration option to choose which one will be used.

[1] https://github.com/systemd/mkosi/commit/99ca14cffb13309a0e6865b2724cc199916c753b
[2] See man kernel-install(8) for details.

OBS PoC: https://build.opensuse.org/project/show/home:afeijoo:poc

### DIFF
--- a/regenerate-initrd-posttrans
+++ b/regenerate-initrd-posttrans
@@ -15,14 +15,29 @@
 # to have all initrds generated. This script is called from posttrans
 # and takes care of generating the initrds
 
-: ${DRACUT:=/usr/bin/dracut}
-if [ ! -x "$DRACUT" ]; then
-    echo "${0##*/}: dracut is not installed, not rebuilding the initrd" >&2
-    exit 0
-fi
+# get the configured INITRD_GENERATOR from /etc/sysconfig/bootloader
+INITRD_GENERATOR="$(. /etc/sysconfig/bootloader 2>/dev/null && echo "$INITRD_GENERATOR")"
+[ "$INITRD_GENERATOR" = "mkosi" ] && INITRD_GENERATOR="mkosi-initrd"
+
+# dracut is the default initrd generator
+: "${INITRD_GENERATOR:=dracut}"
+
+# check if the configured initrd generator is supported
+[ "$INITRD_GENERATOR" = "dracut" ] || [ "$INITRD_GENERATOR" = "mkosi-initrd" ] || {
+	echo "${0##*/}: the initrd generator \"$INITRD_GENERATOR\" configured in /etc/sysconfig/bootloader is not supported." >&2
+	echo "${0##*/}: valid options are \"dracut\" or \"mkosi-initrd\"." >&2
+	exit 1
+}
+
+# check if the specific posttrans script is available
+[ -x "/usr/lib/module-init-tools/${INITRD_GENERATOR}-posttrans" ] || {
+	echo "${0##*/}: missing required /usr/lib/module-init-tools/${INITRD_GENERATOR}-posttrans." >&2
+	exit 1
+}
+. /usr/lib/module-init-tools/"${INITRD_GENERATOR}"-posttrans
 
 if [ -e "/usr/bin/sdbootutil" ] && /usr/bin/sdbootutil is-installed; then
-    is_sdbootutil=1
+	is_sdbootutil=1
 fi
 
 dir=/run/regenerate-initrd
@@ -45,12 +60,12 @@ fi
 
 for f in "$dir"/*; do
 	case $f in
-	    "$dir/*")
+		"$dir/*")
 		[ -e "$f" ] || break;;
 	esac
 	# check if we are in a build chroot
-	if ! [  -f /etc/fstab -a ! -e /.buildenv -a -x "$DRACUT" ] ; then
-		echo "Please run \"$DRACUT -f --regenerate-all\" as soon as your system is complete." >&2
+	if [ ! -f /etc/fstab ] || [ -e /.buildenv ]; then
+		initrd_warn_chroot_build
 		rm "$dir"/*
 		exit 0
 	fi
@@ -66,7 +81,7 @@ if test -e "$dir/all"; then
 		if [ -n "$is_sdbootutil" ]; then
 			/usr/bin/sdbootutil --no-reuse-initrd --default-snapshot add-all-kernels
 		else
-			"$DRACUT" -f --regenerate-all
+			initrd_regenerate_all
 		fi
 		work_done=yes
 	}
@@ -88,18 +103,19 @@ else
 			continue
 		}
 		if [ -n "$is_sdbootutil" ]; then
-			if ! /usr/bin/sdbootutil --no-reuse-initrd --default-snapshot add-kernel "$kver"; then
-				err=$?
-			else
-				work_done=yes
-			fi
-		elif ! "$DRACUT" -f --kver "$kver"; then
+			/usr/bin/sdbootutil --no-reuse-initrd --default-snapshot add-kernel "$kver"
 			err=$?
+			[ $err -eq 0 ] && work_done=yes
 		else
-			work_done=yes
+			initrd_regenerate "$kver"
+			err=$?
+			[ $err -eq 0 ] && work_done=yes
 		fi
 	done
 fi
+
+# Clean-up before exit
+trap 'initrd_cleanup' EXIT
 
 # For XEN/grub2 configurations, make sure the updated initrds are copied
 # to the EFI system partition. See /etc/grub.d/20_linux_xen.
--- a/weak-modules2
+++ b/weak-modules2
@@ -56,7 +56,9 @@
 # kmps: list of kmps, newest first
 #
 
-: ${DRACUT:=/usr/bin/dracut}
+: "${DRACUT:=/usr/bin/dracut}"
+: "${MKOSI_INITRD:=/usr/bin/mkosi-initrd}"
+
 find_lsinitrd() {
     local lsi
     LSINITRD=
@@ -98,6 +100,24 @@ find_sdbootutil() {
 	SDBOOTUTIL="/usr/bin/sdbootutil"
     fi
     dlog "SDBOOTUTIL=$SDBOOTUTIL"
+}
+
+find_initrd_generator() {
+    INITRD_GENERATOR="$(. /etc/sysconfig/bootloader 2>/dev/null && echo "$INITRD_GENERATOR")"
+
+    # dracut is the default initrd generator
+    if [ -z "$INITRD_GENERATOR" ]; then
+	INITRD_GENERATOR="dracut"
+    elif [ "$INITRD_GENERATOR" = "mkosi" ]; then
+	INITRD_GENERATOR="mkosi-initrd"
+    fi
+
+    [ "$INITRD_GENERATOR" = "dracut" ] || [ "$INITRD_GENERATOR" = "mkosi-initrd" ] || {
+	echo "ERROR: the initrd generator \"$INITRD_GENERATOR\" configured in /etc/sysconfig/bootloader is not supported." >&2
+	exit 1
+    }
+
+    dlog "INITRD_GENERATOR=$INITRD_GENERATOR"
 }
 
 find_usrmerge_boot() {
@@ -453,7 +473,7 @@ DRACUT_CONFDIR=/etc/dracut.conf.d
 DRACUT_BUILTIN_CONFDIR=/usr/lib/dracut/dracut.conf.d
 GET_DRACUT_DRIVERS=/usr/lib/module-init-tools/get_dracut_drivers
 
-get_initrd_basenames() {
+get_dracut_basenames() {
     local setpriv=$(command -v setpriv)
     local conf= cf
 
@@ -479,6 +499,24 @@ $(cat "$cf")"
 		   "$GET_DRACUT_DRIVERS" "$conf"
 }
 
+get_mkosi_initrd_basenames() {
+    # TODO: get drivers from mkosi-initrd conf (KernelModulesInclude=, KernelModulesExclude=)
+    get_current_basenames "$1"
+}
+
+get_initrd_basenames() {
+    local krel=$1
+
+    case "$INITRD_GENERATOR" in
+	"dracut")
+	    get_dracut_basenames "$krel"
+	    ;;
+	"mkosi-initrd")
+	    get_mkosi_initrd_basenames "$krel"
+	    ;;
+    esac
+}
+
 # test if rebuilding initrd is needed for $krel.
 # stdin - list of changed modules ("_kernel_" for the whole kernel)
 needs_initrd() {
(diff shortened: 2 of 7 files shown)


regenerate-initrd-posttrans
```diff
@@ -15,14 +15,84 @@
...
+				"/usr/local/lib/kernel/install.conf.d/$conf" \
+				"/usr/lib/kernel/install.conf.d/$conf" \
+				; do
+				[ -f "$f" ] || continue
+				. "$f"
+				if [ -n "$initrd_generator" ]; then
+					echo "$initrd_generator"
+					break 2
+				fi
+			done
```
mwilck, line 50: Shouldn't you `break` here too? Otherwise the same conf file can be read from 2 directories

regenerate-initrd-posttrans
```diff
@@ -15,14 +15,84 @@
...
 
+# Check if `initrd_generator` is defined. See kernel-install(8) for details.
+# Note on how systemd gets this config:
+# - All the dropins are included in alphabetical order, no matter the location.
+# - If the variable is defined in any dropin, the main config file is ignored,
+# so in that case we don't need to find it.
+# - It only reads the first main config file (kernel/install.conf). From highest
+# to lowest priority: /etc, /usr/local/lib, /usr/lib.
+get_kernel_install_dropins_conf() {
+	kernel_install_dropins_conf=
```
mwilck, line 26: Use `local` ?

regenerate-initrd-posttrans
```diff
@@ -15,14 +15,84 @@
...
+					echo "$initrd_generator"
+					break 2
+				fi
+			done
+		done
+	fi
+}
+
+initrd_generator="$(get_kernel_install_dropins_conf)"
+if [ -z "$initrd_generator" ]; then
```
mwilck, line 56: I suggest moving this part into `get_kernel_install_dropins_conf` (and possibly rename the function).

regenerate-initrd-posttrans
```diff
@@ -15,14 +15,84 @@
...
-if [ ! -x "$DRACUT" ]; then
-    echo "${0##*/}: dracut is not installed, not rebuilding the initrd" >&2
-    exit 0
+if [ "$initrd_generator" = "dracut" ] && [ ! -x "$DRACUT" ]; then
+	echo "${0##*/}: dracut is not installed, not rebuilding the initrd" >&2
+	exit 0
+fi
+
+: ${MKOSI_INITRD:=/usr/bin/mkosi-initrd}
+[ "$initrd_generator" = "mkosi" ] && initrd_generator="mkosi-initrd"
```
mwilck, line 79: This will error out if the script is ever run with `bash -e`. Please use `||` or `if`.

mwilck: I suppose we *should* error out if `$initrd_generator` is neither `dracut` nor `mkosi`, but in that case we should print a meaningful error message.

regenerate-initrd-posttrans
```diff
@@ -45,12 +115,21 @@ fi
...
-		echo "Please run \"$DRACUT -f --regenerate-all\" as soon as your system is complete." >&2
+	if ! [  -f /etc/fstab -a ! -e /.buildenv ]; then
+		case "$initrd_generator" in
+			"dracut")
+				echo "Please run \"$DRACUT -f --regenerate-all\" as soon as your system is complete." >&2
+				;;
+			"mkosi-initrd")
+				# FIXME: mkosi-initrd does not provide anything like --regenerate-all yet
+				echo "Please regenerate all the initrds with \"$MKOSI_INITRD\" as soon as your system is complete." >&2
+				echo "WARNING: mkosi-initrd cannot generate the initrd directly into /boot, you need to use an intermediate directory." >&2
```
mwilck, line 130: Not sure if this warning is necessary; people using `mkosi-initrd` should know what they're doing.

regenerate-initrd-posttrans
```diff
@@ -93,14 +189,37 @@ else
...
+						work_done=yes
+					fi
+					;;
+			esac
 		fi
 	done
 fi
 
+# Clean-up before exit
+if [ "$initrd_generator" = "mkosi-initrd" ]; then
```
mwilck, line 219: Please use an exit trap here (`trap 0 'rm -rf "$MKOSI_INITRD_STAGING_DIR"'`).

regenerate-initrd-posttrans
```diff
@@ -93,14 +189,37 @@ else
...
+					;;
+				"mkosi-initrd")
+					if ! "$MKOSI_INITRD" --kernel-version "$kver" -O "$MKOSI_INITRD_STAGING_DIR" -o "initrd-$kver"; then
+						err=$?
+					elif ! chmod 600 "$MKOSI_INITRD_STAGING_DIR/initrd-$kver"; then
+						err=$?
+					elif ! cp --reflink=auto "$MKOSI_INITRD_STAGING_DIR/initrd-$kver" "/boot/initrd-$kver"; then
+						err=$?
+					else
+						rm -f "$MKOSI_INITRD_STAGING_DIR/initrd-$kver"*
```
mwilck, line 209: You don't need to do this if you remove the entire staging directory.

regenerate-initrd-posttrans
```diff
@@ -93,14 +189,37 @@ else
...
+					else
+						work_done=yes
+					fi
+					;;
+				"mkosi-initrd")
+					if ! "$MKOSI_INITRD" --kernel-version "$kver" -O "$MKOSI_INITRD_STAGING_DIR" -o "initrd-$kver"; then
+						err=$?
+					elif ! chmod 600 "$MKOSI_INITRD_STAGING_DIR/initrd-$kver"; then
+						err=$?
+					elif ! cp --reflink=auto "$MKOSI_INITRD_STAGING_DIR/initrd-$kver" "/boot/initrd-$kver"; then
```
mwilck, line 206: Is `--reflink=auto` portable? It's the default anyway for GNU cp, no?

weak-modules2
```diff
@@ -100,6 +102,75 @@ find_sdbootutil() {
...
+				"/etc/kernel/install.conf.d/$conf" \
+				"/usr/local/lib/kernel/install.conf.d/$conf" \
+				"/usr/lib/kernel/install.conf.d/$conf" \
+				; do
+				[ -f "$f" ] || continue
+				. "$f"
+				if [ -n "$initrd_generator" ]; then
+					echo "$initrd_generator"
+					break 2
+				fi
```
mwilck, line 138: like above, need a `break` here IMO

regenerate-initrd-posttrans
mwilck: `local` isn't POSIX but all shells we ship support it, AFAIK.

weak-modules2
```diff
@@ -100,6 +102,75 @@ find_sdbootutil() {
...
+			fi
+		done
+	fi
+
+	# dracut is the default initrd generator
+	if [ -z "$INITRD_GENERATOR" ]; then
+		INITRD_GENERATOR="dracut"
+	elif [ "$INITRD_GENERATOR" = "mkosi" ]; then
+		INITRD_GENERATOR="mkosi-initrd"
+	fi
```
mwilck, line 169: like above, error out if it's neither?

weak-modules2
```diff
@@ -457,26 +528,35 @@ get_initrd_basenames() {
...
-    [[ ! -f "$DRACUT_CONFFILE" ]] ||
-	conf="$(cat "$DRACUT_CONFFILE")"
-
-    # Assemble the content of the dracut configuration files.
-    # Make sure to put a newline between subsequent conf file contents.
-    for cf in $(dropindirs_sort .conf "$DRACUT_CONFDIR" "$DRACUT_BUILTIN_CONFDIR"); do
-	conf="$conf
+	case "$INITRD_GENERATOR" in
+		"dracut")
+			[[ -x "$GET_DRACUT_DRIVERS" && "$setpriv" ]] || {
```
mwilck, line 533: Quite large indentation—move this into a separate function?

mwilck (changes requested):
Looking good overall, some minor comments.

mwilck:
@aafeijoo-suse, I realize that you are leveraging features from `kernel-install`.

I wonder if you're aware of https://jira.suse.com/browse/PED-7184 and https://bugzilla.suse.com/show_bug.cgi?id=1194723, where support for `kernel-install` has been discussed, but little progress has been made so far.

aafeijoo-suse:
> @aafeijoo-suse, I realize that you are leveraging features from `kernel-install`.
> 
> I wonder if you're aware of https://jira.suse.com/browse/PED-7184 and https://bugzilla.suse.com/show_bug.cgi?id=1194723, where support for `kernel-install` has been discussed, but little progress has been made so far.

Thanks for the pointer, I wasn't aware of that discussion, will take a look today. I just found useful the existing `initrd_generator=` config option (instead of creating our own custom option in `/etc/sysconfig/xxx`?). Since we use systemd, `kernel-install` is always shipped, and also grepping `kernel-install inspect` allows to get the configured initrd generator. But yes, I'm not 100% sure if we should use `kernel-install` configuration options if `kernel-install` does not work well for us.

aafeijoo-suse:
> > @aafeijoo-suse, I realize that you are leveraging features from `kernel-install`.
> > I wonder if you're aware of https://jira.suse.com/browse/PED-7184 and https://bugzilla.suse.com/show_bug.cgi?id=1194723, where support for `kernel-install` has been discussed, but little progress has been made so far.
> 
> Thanks for the pointer, I wasn't aware of that discussion, will take a look today. I just found useful the existing `initrd_generator=` config option (instead of creating our own custom option in `/etc/sysconfig/xxx`?). Since we use systemd, `kernel-install` is always shipped, and also grepping `kernel-install inspect` allows to get the configured initrd generator. But yes, I'm not 100% sure if we should use `kernel-install` configuration options if `kernel-install` does not work well for us.

Definitely it does not make sense the current approach. What do you think about adding `INITRD_GENERATOR=` to `/etc/sysconfig/bootloader`?

regenerate-initrd-posttrans
keszybz: > This will error out if the script is ever run with `bash -e`. Please use `||` or `if`.

Actually, no. With `&&` (or `||`), it's a conditional expression, and failure of those expressions does not cause termination.

mwilck: I stand corrected. This is embarrassing. I could swear I'd observed the behavior I falsely described many times ... but maybe I've been fooled by something else, can't tell any more.

Thanks for pointing it out.

mwilck:
> Definitely it does not make sense the current approach. What do you think about adding `INITRD_GENERATOR=` to `/etc/sysconfig/bootloader`?

OK for me, personally. But I have heard from @thkukuk and @lnussel that `/etc/sysconfig` is being (or should be) phased out as well.

regenerate-initrd-posttrans
keszybz: :)

thkukuk:
> > Definitely it does not make sense the current approach. What do you think about adding `INITRD_GENERATOR=` to `/etc/sysconfig/bootloader`?
> 
> OK for me, personally. But I have heard from @thkukuk and @lnussel that `/etc/sysconfig` is being (or should be) phased out as well.

We should not introduce new ones, but `/etc/sysconfig` is very limited if we as distributor needs to make changes to it or if it is no longer needed. We only know found `/etc/sysconfig` files and variables, which were unused since 7 years. But everybody wasted a lot of time maintaining them...

In this case: we have already `/etc/sysconfig/bootloader`, continue to use it.

About `kernel-install`: do we use that? The last time Ludwig looked at it, it was pretty unuseable for our needs and upstream not willing to accept the missing hooks. But we need to make sure this works with `sdbootutil`. @aplanas?

aplanas:
> In this case: we have already `/etc/sysconfig/bootloader`, continue to use it.

One comment from my side: `/etc/sysconfig/bootloader` does not have owner. Is a file created by YaST and unconditionally used by perl-Bootloader, but they do not own it.  This means that new images will lack this file and its content.

> About `kernel-install`: do we use that? The last time Ludwig looked at it, it was pretty unuseable for our needs and upstream not willing to accept the missing hooks. But we need to make sure this works with `sdbootutil`. @aplanas?

Yes please, for now use `sdbootutil add-kernel` (https://github.com/openSUSE/suse-module-tools/blob/master/kernel-scriptlets/rpm-script#L109). One reason is that it will take care of details that are relevant for MicroOS. To generate an initrd you can use `sdbootutil mkinitrd` that is equivalent to `sdbootutil add-kernel --no-reuse-initrd`

mwilck:
> About `kernel-install`: do we use that?

We currently don't, but there have been requests that we do.
@thkukuk, I'd be grateful if you could join the discussion on PED-7184.

aafeijoo-suse:
Thanks for your feedback. Reworked to parse `/etc/sysconfig/bootloader` instead. Also updated and tested the OBS project: https://build.opensuse.org/project/show/home:afeijoo:poc

PTAL.

regenerate-initrd-posttrans
aafeijoo-suse: dracut explicitly uses `--reflink=auto` everywhere since forever, I didn't check if it's the default now. I can remove it if you want, but it doesn't hurt IMO.

mwilck: Ok then. Just verified that `busybox cp` at least doesn't bail out on this option.

mwilck:
@aafeijoo-suse, is this ready for merging now?

aafeijoo-suse:
> @aafeijoo-suse, is this ready for merging now?

It's from my side.

mwilck (changes requested):
There are still some rough edges here.

Actually for regenerate-initrd-posttrans, please consider moving the dracut and mkosi specific code into two different executables (e.g. `dracut-posttrans` and `mkosi-posttrans`), and converting r-i-p itself into a tiny driver script.

aafeijoo-suse:
> Actually for regenerate-initrd-posttrans, please consider moving the dracut and mkosi specific code into two different executables (e.g. `dracut-posttrans` and `mkosi-posttrans`), and converting r-i-p itself into a tiny driver script.

Do you actually want 2 different executables duplicating most of the common code? or just 2 different scripts with the same function names, to source one or another depending on the initrd context?

mwilck:
> Do you actually want 2 different executables duplicating most of the common code? or just 2 different scripts with the same function names, to source one or another depending on the initrd context?

I had the impression that there was not much common code left. Perhaps I was wrong. I didn't like the repeated case statements. Sourcing function scripts might be good idea.

aafeijoo-suse:
> I had the impression that there was not much common code left. Perhaps I was wrong. Sourcing function scripts might be good idea.

Proposed changes are ready for review, PTAL. I did some tests on both transactional and non-transactional systems using dracut and mkosi-initrd, apparently everything looks good. OBS PoC updated as well. Thanks.

mwilck (approved):
LGTM

aafeijoo-suse:
Thanks!

PR #107, opened 2024-08-09
Requires sdbootutil if already installed

aplanas:
If sdbootutil is used in the scriptlets, we should require it.  This can fix installation ordering issues like bsc#1228759 or bsc#1228659.


mwilck:
LGTM, merged into master

PR #108, opened 2024-08-29
Generate initrd for the default snapshot

aplanas:
In transactional systems, regenerate-initrd-posttrans script is called outside the transaction.  In this case the current snapshot is not the default one, and we need to generate the initrd for the default one.

This patch calls sdbootutil with the --default-snapshot parameter, to refer to the correct snapshot.

### DIFF
--- a/regenerate-initrd-posttrans
+++ b/regenerate-initrd-posttrans
@@ -64,7 +64,7 @@ if test -e "$dir/all"; then
 	rm "$dir"/*
 	[ "$SKIP_REGENERATE_INITRD_ALL" = 1 ] || {
 		if [ -n "$is_sdbootutil" ]; then
-			/usr/bin/sdbootutil --no-reuse-initrd add-all-kernels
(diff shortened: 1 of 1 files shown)


aplanas:
Fixes bsc#1224773

aplanas:
Depends on https://github.com/openSUSE/sdbootutil/pull/126

mwilck:
So `--default-snapshot` does The Right Thing™ for both transactional and non-transactional systems?

aplanas:
> So `--default-snapshot` does The Right Thing™ for both transactional and non-transactional systems?

That is correct, it will point to the default transaction in both cases. In a normal Tumbleweed the default and the active ones are the same, but in MicroOS they will be different.

But should be noted that this does not matter, as if sdbootutil is installed, this script will always be executed outside the transaction: https://github.com/openSUSE/suse-module-tools/pull/108/files#diff-a5683575be1701b9948a3b07c0824f986dd168dc68e64e18110d21fc5a4606c2L34-L44

PR #109, opened 2024-09-03
Fix handling of /boot/initrd and /boot/vmlinuz

mwilck:
- **rpm-script: fix handling of /boot/vmlinuz and /boot/initrd**
- **rpm-script: remove /boot/vmlinuz and /boot/initrd on kernel removal**

See [bsc#1207703](https://bugzilla.suse.com/show_bug.cgi?id=1207703)

### DIFF
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -133,10 +133,10 @@ update_bootloader_entry() {
 		    message_install_bl
 		fi
 	    fi
-	    create_boot_symlinks
+	    create_boot_symlinks "$kernelrelease-$flavor"
 	else
 	    message_install_bl
-	    create_boot_symlinks
+	    create_boot_symlinks "$kernelrelease-$flavor"
 	fi
     fi
 }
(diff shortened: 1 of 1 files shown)


mwilck:
Also @aafeijoo-suse

hramrach:
The remove on kernel removal part is bad. If anything at all relies on the symlink to be usable removing it breaks that.

Also it is debatable if the last installed kernel or latest kernel should provide the symlink.

Looks like we would need full update-alternatives or similar to get this correct.

mwilck:
> The remove on kernel removal part is bad. If anything at all relies on the symlink to be usable removing it breaks that.

I only remove the symlinks if they are stale, or will become stale immediately after the package removal. What purpose would a stale symlink serve?

> Also it is debatable if the last installed kernel or latest kernel should provide the symlink.

"latest installed" has always been the way it has worked. That makes sense, IMO, and I see no reason to change it. AFAICS, it's consistent with what the `bootloader_entry` script does.

> Looks like we would need full update-alternatives or similar to get this correct.

With the current efforts to integrate sd-boot, this is going to be the "legacy" code path soonish.

The only case that needs fixing is when the kernel currently owning the symlink gets removed.

We could delegate the symlink generation to the `bootloader_entry` script. That would make sure that the default entry and the symlinks are always consistent, even in the "remove" case.

Alternatively, we could fix the removal by some simple shell logic selecting the "latest" of the installed kernels, if any. 

I'd like to postpone either solution to a separate patch set, though.

hramrach:
A stale symlink serves as much purpose as none. It does not point to a bootable kernel.

If it's required that such symlink exists then it's still broken.

Sure, this does fix some cases.

It does not, however, provide a symlink that always points somewhere.

If that is required this is not a full solution.

If it is not required and everything is best-effort then arguably the current code is also good enough.

mwilck:
Hm again, the following isn't true:

> AFAICS, it's consistent with what the bootloader_entry script does.

perl-Bootloader simply calls `grub2-mkconfig`, which uses a complex algorithm based on version comparison to determine the kernel to use in the topmost entry. Trying to mimic this logic is definitely out of scope for `rpm-script`.

hramrach:
It should now have some argument to give the 'top' kernel, used by kexec-bootloader.

Won't help if the 'top' kernel happens to be the one with no ramdisk, though.

mwilck:
I pushed another patch that implements "best effort" handling for the symlinks in case of kernel package removal.

mwilck:
Right, there is `pbl --default-settings`.

mwilck:
Haha, this was a waste of time. `pbl --default-settings` is basically nothing more than `readlink /boot/initrd`. It does _not_ parse `grub.cfg`.

mwilck:
The current code in this PR is the closest we can get with "best effort". @hramrach, @aafeijoo-suse, please have another look.

hramrach:
So pbl relies on the symlink, that's 'great' design.

kernel-scriptlets/rpm-script
```diff
@@ -330,6 +343,14 @@ case $op in
 			 [ "$(readlink /boot/initrd)" = "initrd-$kernelrelease-$flavor" ]; }; then
 		rm -f /boot/initrd "/boot/$image"
 	    fi
+	    if [ ! -f /boot/initrd ] || [ ! -f "/boot/$image"; then
```
aafeijoo-suse, line 346: If I'm not mistaken you have a syntax error here, missing `]`

mwilck: ups

aafeijoo-suse:
IMHO this improves the broken-symlink situation (at a first glance, I haven't tested it).
Regarding bsc#1207703, I thought that `initrd` on standard GRUB menu entries use to point to `/boot/initrd-<kver>`, not to this symlink. Maybe that was the case on very ancient systems?

mwilck:
> Regarding bsc#1207703, I thought that initrd on standard GRUB menu entries use to point to /boot/initrd-<kver>, not to this symlink.

True.

> Maybe that was the case on very ancient systems?

I have no idea, but probably yes.

mwilck:
Objections against merging this?

aafeijoo-suse:
> Objections against merging this?

Not from my side.

mwilck:
@hramrach has expressed some general doubts but didn't say anything against it. I'm merging this now.

PR #110, opened 2024-09-20
rpm-script: create vmlinuz and initrd also in image build environments

mwilck:
See boo#1230858.


aplanas (commented):
LGTM but I did not test it

PR #111, opened 2024-10-21
udevrules: Add rule to set maximum readahead window

jankara:
Add udev rule to allow tuning the maximum readahead window for the devices.


mwilck:
This not only allows tuning, it also increases the default. I suppose this is intentional, but shouldn't we mention it in the commit, and perhaps even in our release notes?

jankara:
Right. The change is indeed intentional. I did some benchmarking and shared the results here: https://jira.suse.com/browse/PED-11154. I'm not sure if it makes sense to put this link into commit message but probably I can put there the gist of it.

jankara:
I've updated the commit and force-pushed it.

PR #114, opened 2025-02-05
Allow KMP metapackages

hramrach:


mwilck:
What are "KMP metapackages"? Can you give me an example?

mwilck:
I still don't understand what you're trying to achieve here.

PR #115, opened 2025-02-14
mkosi-initrd: build initrds directly to /boot

aafeijoo-suse:
Before v25, mkosi-initrd failed to build initrds directly to /boot, and also to set proper permissions. Even the requested output file was a symlink to the compressed file with another extension. With v25 in Factory since snapshot 20250124, some workarounds can be removed.

Ref:
- https://github.com/systemd/mkosi/commit/3454f7bd4ef0336ec80a117d593baaef0fe53398
- https://github.com/systemd/mkosi/commit/3fe62ba52099e38920dc07cf3d3a90d048dfa954
- https://github.com/systemd/mkosi/commit/6b0dfe58f3f04264f1df5cb90b7091195913562f

### DIFF
--- a/dracut-posttrans
+++ b/dracut-posttrans
@@ -19,7 +19,3 @@ initrd_regenerate() {
 initrd_regenerate_all() {
 	"$DRACUT" -f --regenerate-all
 }
-
-initrd_cleanup() {
-	:
-}
--- a/mkosi-initrd-posttrans
+++ b/mkosi-initrd-posttrans
@@ -7,13 +7,6 @@ if [ ! -x "$MKOSI_INITRD" ]; then
 	exit 0
 fi
 
-# FIXME: mkosi-initrd fails to cp files directly to /boot
-MKOSI_INITRD_STAGING_DIR="$(mktemp -p /var/tmp/ -d -t mkosi-initrd-stagingXXXXXXXX)"
-if [ ! -d "$MKOSI_INITRD_STAGING_DIR" ]; then
-	echo "${0##*/}: failed to create mkosi-initrd staging directory." >&2
-	exit 1
-fi
-
 initrd_warn_chroot_build() {
 	# FIXME: mkosi-initrd does not provide anything like --regenerate-all yet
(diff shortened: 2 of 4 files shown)


aafeijoo-suse:
> Does this pose additiional risk for the build procedure to fail, in particular when an initrd image of the same name exists already?

No, if a file with the same name already exists, it is overwritten.

> Does mkosi write directly to the configured output file, or does it write to a temporary file and rename atomically after building successfully?

Now it uses internally a temporary directory to build the image, then it copies the image to the specified output dir + file. 


FYI, you can test the patch here: https://build.opensuse.org/package/show/home:afeijoo:branches:openSUSE:Factory:DEBUG/suse-module-tools

mwilck:
> No, if a file with the same name already exists, it is overwritten.

This is actually a concern. the new image is larger than the old one and space is scarce, the (good) existing image may be overwritten with something that's incomplete.

But this is the same for dracut and should probably addressed in the future.

aafeijoo-suse:
> > No, if a file with the same name already exists, it is overwritten.
> 
> This is actually a concern. the new image is larger than the old one and space is scarce, the (good) existing image may be overwritten with something that's incomplete.

Yes, both end up using `cp`, and it can fail with `No space left on device` leaving an incomplete initrd...

> But this is the same for dracut and should probably addressed in the future.

This is actually something that needs to be fixed, I'll submit a patch upstream.

Thanks!

aafeijoo-suse:
> > > No, if a file with the same name already exists, it is overwritten.
> > 
> > 
> > This is actually a concern. the new image is larger than the old one and space is scarce, the (good) existing image may be overwritten with something that's incomplete.
> 
> Yes, both end up using `cp`, and it can fail with `No space left on device` leaving an incomplete initrd...
> 
> > But this is the same for dracut and should probably addressed in the future.
> 
> This is actually something that needs to be fixed, I'll submit a patch upstream.

FTR:
- https://github.com/systemd/mkosi/pull/3589
- https://github.com/dracut-ng/dracut-ng/pull/1243

PR #116, opened 2025-04-22
mkosi-initrd: perform some basic checks on regenerate all

aafeijoo-suse:
dracut performs some basic checks before building the initrds with `--regenerate-all` for all kernel directories, avoiding the build without returning error if the checks are not met. So we have to do the same for mkosi-initrd. This fixes system updates when KMPs are installed, as they install their kernel modules in the updates folder of a specific kernel directory, even if that directory has no other content.


aafeijoo-suse:
> Sorry, this had fallen through the cracks.

No worries.

> Do we need it in any older code streams?

No, only for Tumbleweed. Thanks.

PR #118, opened 2025-05-09
Avoid requiring util-linux and awk

mwilck:
suse-module-tools uses tools from util-linux and awk. The required functionality can be easily implemented with coreutils and the shell itself though-

### DIFF
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -169,11 +169,23 @@ check_space_in_boot() {
     fi
 }
 
+usr_or_boot_is_mountpoint() {
+    cut -d" " -f 5 </proc/self/mountinfo | { \
+	while read -r _mp; do
+	    case $_mp in
+		/boot|/usr)
+		    return 0;;
+	    esac
+	done
+	return 1
+    }
+}
+
 copy_or_link_legacy_files() {
     # compat stuff for /boot.
     # if /boot and /usr are not separate partitions we can just link
     # the kernel there to save space. Otherwise copy.
-    if mountpoint -q /boot || mountpoint -q /usr; then
+    if usr_or_boot_is_mountpoint; then
 	copy_or_link="cp -a --remove-destination"
 	separate_boot='1'
     else
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -1,4 +1,5 @@
 #!/bin/sh
+[ ! "$KERNEL_PACKAGE_SCRIPT_TRACE" ] || set -x
 
 op=${0##*-}
 
--- a/kernel-scriptlets/rpm-script
+++ b/kernel-scriptlets/rpm-script
@@ -155,7 +155,8 @@ check_space_in_boot() {
     # plugin (that calls sdbootutil add-all-kernels)
     #
     if [ "$YAST_IS_RUNNING" != "" ]; then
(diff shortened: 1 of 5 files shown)


kernel-scriptlets/rpm-script
```diff
@@ -155,7 +156,8 @@ check_space_in_boot() {
     # plugin (that calls sdbootutil add-all-kernels)
     #
     if [ "$YAST_IS_RUNNING" != "" ]; then
-	mydf="$( POSIXLY_CORRECT=1 df -P /boot/ | awk '/^(\/|-[[:blank:]])/{ print $4}' )"
+	mydf="$( POSIXLY_CORRECT=1 df -P /boot/ 2>/dev/null | tail -n 1 | \
+		 while read -r _m _t _u _a _r; do echo "$_a"; done )"
```
hramrach, line 160: sounds like job for `cut`

mwilck:
Thanks for pointing these out, @aafeijoo-suse !

> https://github.com/openSUSE/suse-module-tools/blob/cf71cc481efe9a715bac2789d91e666f98cb743a/weak-modules2#L476

this one is non-fatal.

> https://github.com/openSUSE/suse-module-tools/blob/cf71cc481efe9a715bac2789d91e666f98cb743a/driver-check.sh#L261

driver-check is basically obsolete. I wouldn't add a package requirement for it.

> https://github.com/openSUSE/suse-module-tools/blob/cf71cc481efe9a715bac2789d91e666f98cb743a/udevrules/80-hotplug-cpu-mem.rules#L26

So "mount" is part of `util-linux`? Not quite getting how we got bug reports from systems where kernel packages are installed but "mount" is not available ... Containers? 

Anyway, I doubt that this udev rule would be executed on any system that doesn't have "mount" installed.

mwilck:
After discussion with @thkukuk, it has become clear that util-linux is actually a mandatory package under (open)SUSE and that the problems were observed with the upstream systemd package only, which is missing this dependency.

The dependency will be added.

aafeijoo-suse:
> After discussion with @thkukuk, it has become clear that util-linux is actually a mandatory package under (open)SUSE and that the problems were observed with the upstream systemd package only, which is missing this dependency.

That is not 100% correct. It can be reproduced with image builders where you can choose what packages you want (e.g. mkosi) . The problem is only about package ordering. Before replacing `Requires: dracut` with `OrderWithRequires: dracut` on the kernel-default package, dracut was installed before it, and util-linux as well because dracut requires it. But now dracut is kind of "optional" after this change, so it's not being installed on this minimal images because no one is requiring it, and util-linux (although is being installed) is not ordered before the kernel package.

Before:

```
2025-04-10T08:42:24.2839397Z (322/418) Installing: util-linux-2.40.4-4.2.x86_64 [..done]
...
2025-04-10T08:42:29.1309491Z (407/418) Installing: dracut-059+suse.714.gee616a76-1.1.x86_64 [..done]
...
2025-04-10T08:42:42.6922551Z (416/418) Installing: kernel-default-6.14.0-1.1.x86_64 [....................................................................................................
```

After:

```
2025-04-24T06:12:16.5211892Z (321/409) Installing: kernel-default-6.14.2-1.1.x86_64 [....................................................................................................
...
2025-04-24T06:12:16.7725336Z (323/409) Installing: util-linux-2.40.4-4.2.x86_64 [..done]
```

> The dependency will be added.

Thanks Martin!

PR #121, opened 2025-09-22
Sles12 sp7 alternative kmp path

e4t:
This is experimental code that adds /lib/modules/kmp(/<kver>) als alternativen module install path. The advantage of this path is that the KMP is not directly installed into the kernel it was built with. Instead weak-updates2 should take care of making this available for this kernel as well.
This ensures there are not multiple versions of the same module available for a single kernel.

### DIFF
--- a/driver-check.sh
+++ b/driver-check.sh
@@ -176,7 +176,7 @@ check_kmp()
 	if ! rpm -q -R "$kmp" | grep -Eq "$req_re"; then
 		error "$kmp does not have proper dependencies"
 	fi
-	exec 3< <(sed -rn 's:^(/lib/modules)?/([^/]*)/(.*\.ko(\.[gx]z|\.zst)?)$:\1 \2 \3:p' \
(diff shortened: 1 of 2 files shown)


hramrach:
Why does the /kmp/ need to be added?

Shouldn't modules installed in /usr/lib/modules/nvdia-G04-175.25 already get scanned for modules?

If not why not?

hramrach:
Duplicate of #123

e4t:
> Why does the /kmp/ need to be added?
> 
> Shouldn't modules installed in /usr/lib/modules/nvdia-G04-175.25 already get scanned for modules?
> 
> If not why not?

With `multiversion(kernel)` set but **without** `(/usr)/lib/modules/kmp` we will have multiple modules installed (not 'weak-update' linked!) for the build kernel (ie. the kernel the KMP is built against). Thus, we will leave it to `depmod` to pick the right module.
If we install KMPs outside of the kernel (ie. in `(/usr)/lib/modules/kmp` ) we leave it to `weak_modules2` to pick one kernel and create a `weak-update` link - giving the control to it instead of relying on depmod to do the right thing.
This is one step towards making `multiversion(kernel)` work. I'm not a fan of it and I'm advocating to drop it from the NVIDIA open driver, but since others seem to find it useful we should make it work - and I happened to have the patch already.

PR #123, opened 2025-09-29
Sles15 sp7 alternative module path

e4t:
This pull request contains:
1. The changes required to support a separate KMP directory in `/lib/modules/kmp`. These changes may not be sufficient.
2. The changes required to find all potential update candidates even if the KMPs have 'versioned' module directory to avoid file conflicts whith `multiversion` support enabled.
3. Changes required when merging 1. and 2. together.
This requires testing as pieces may be missing.


hramrach:
Why does the /kmp/ need to be added?

Shouldn't modules installed in /usr/lib/modules/nvdia-G04-175.25 already get scanned for modules?

If not why not?

hramrach:
I can't read your mind.  I do not know what is in that /kmp directory.

So your design is to prepend /kmp to the kernel version when installing KMPs.

And my question is why do that instead of installing under a directory named after the KMP.

And yes, it is not clear why merely prepending /kmp is more likely to work than installing in a completely arbitrary directory under /lib/modules.

hramrach:
Nor in `(/usr)/lib/modules/<kmpname>-<kmpver>`

hramrach:
I see what is wrong with this approach now.

The group that comes after /modules/ and with the change after /modules/kmp/ captures the kernel version.

But the out-of-kernel KMPs do not have a kernel version.

That's why prepending '/kmp' was seen as a 'solution' to make this code 'work' but it does not. The modules that are out of kernel should be versionless.

e4t:
> or `/lib/modules/<kmp_name>-<module_ver>-<kver>`
> 
In other words, the `nvidia_drm.ko.zst` would then reside in: 
`/lib/modules/nvidia-open-driver-G06-signed-570.144-6.4.0-150700.51-default/updates/` - correct?
Why would this be better than: `/lib/modules/kmp/6.4.0-150700.51-default/updates/nvidia-open-driver-G06-signed-570.144` ?

Large portions of the code rely on the fact that the directory structure is:
```
<prefix>/<kver>/updates/<kmp_name>/<module>.ko
```
with this, we will have to support another variant:
```
<prefix>/<module_name>-<kver>/updates/<module>.ko
```
The changes required to support both variants in weak-updates2 would be quite intrusive with a greater risk of breakages.

> The point is that people intentionally installed the nvidia package, they don't care if it was delivered through KMP or DKMS, and very few people actually know what that is. Putting the installation method before what is installed needlessly obfuscates what is installed.

I'm not sure if people actually care.

> DKMS does not actually work correctly on (open)SUSE but before Linux version 6 the nVidia modules were installed through some DKMS-like method without actually using DKMS which needlessly rebuilt the module for every kernel. If it installed the module to a separate directory and used wm2 it could skip most builds, and DKMS could too if it is ever integrated.

NVIDIA's proprietary driver used a home-baked build/install method on SUSE and DKMS everywhere else. As they also provide a package with the open driver, they now us DKMS everywhere - including SUSE. 
DKMS used outside of package install wants to install in a ro file system in case transactional updates are enabled, so we'd have to fix it.
The way DKMS operates is quite different from what we are trying to achieve with multiversioned packages.

What we are are trying to achieve with multiversioned packages is not wrong - and the SolidDriver people seem to like it. It is just very broken in some respects: that we can only publish a single module per kernel release and are thus bound to one release every four weeks is a major user experience problem.
With my patches I was trying to mitigate this flaw. But at the same time, I decided to discontinue multiversion because I knew that changing anything will be a huge bike shed and we will not achieve anything in time while the next NVIDIA driver update is just a day away.
Since I've made these changes already I provided them so they don't get lost and my time has been wasted. If there are flaws in my code, I'm happy to look into them, catering to different perceptions of 'obscurity' is not on my charter. 
If you feel things need to be done differently, feel free to do so. I've got other things on my plate.

e4t:
> Why `/updates/`? `/lib/modules/nvidia-open-driver-G06-signed-570.144-6.4.0-150700.51-default` is its own directory, it is not updating itself.
> 
Ok, it would make it easier to discover what scheme is used - the 'new' one under `(/usr)/lib/modules/<versioned_module_name>-<kver>` or the old one `(/usr)/lib/modules/<kver>`.

> And supporting another variant is warranted, the modules that are installed out of kernel need to be handled differently.

Yes.

> That is the group that captures kernel version should be tightened to reject anything that does not start with at least two place version such as `6.0~some-non-numeric-stuff`, and second match attempt done without capturing the version at all.

I'm not fully sure if I understand, but we should be stricter on the requirements of the module directory. The weak link in the kernel should probably be in a directory `(/usr)/lib/modules/<kver>/updates/<unversioned_module_name>/`as this will avoid duplicate links and will make it easier to identify an update candidate.  For this, we probably should adopt the convention that <unversioned_module_name> is held whereever needed  in a variable module_basename or similar.

e4t:
> I see what is wrong with this approach now.
> 
> The group that comes after /modules/ and with the change after /modules/kmp/ captures the kernel version.
> 
> But the out-of-kernel KMPs do not have a kernel version.
> 

Out-of-kernel KMPs do have a kernel version. It's the version they are built
against. The kernel version determines the ABI - which should be stable for an enterprise product version but may be not.

> That's why prepending '/kmp' was seen as a 'solution' to make this code 'work' but it does not. The modules that are out of kernel should be versionless.

Why?

e4t:
Ok, now I understand what you mean. 
You are trying to solve a much bigger challenge than I did with the original patch. 
You would like to go the full step: by no longer installing a KMP into it's 'native' kernel ('native' referring to the kernel it's been built with), we don't need to associate this module with a particular kernel version as we already rely on the on the ABI checker to determine what kernels a module is compatible with.
This is definitely a valid approach but a much bigger step - not suitable for a maintenance update - at least not without extensive testing and possibly an ECO as we need to alert others - not because it is complicated but because it is error prone. What makes it even more so is that we still need to support the old scheme.
Note on the side: we still need to be able to determine the kernel a KMP has been built with simply for tracability: for example, the KMP may have picked up something from it's 'native' kernel (for instance through a macro) that later gets proven to be problematic.

e4t:
> Why is it moot? You can make the krel group optional, swap `\1 \2 \3` to `\1 \3 \2` and then `prefix krel path` to `prefix path krel` and use the parts that are relevant.

You don't check $krel and your $path test would be different. Just the test for $prefix makes sense unaltered. This is why I suggest to do an entirely separate loop for this case.
But these are implementation details we don't need to discuss piece by piece here.

mwilck:
@hramrach 

> But the out-of-kernel KMPs do not have a kernel version.
> 
> That's why prepending '/kmp' was seen as a 'solution' to make this code 'work' but it does not. The modules that are out of kernel should be versionless.

I am not getting this argument. As @e4t wrote, of course they do. It's part of their package version.

You're right that if we install the modules outside the tree of the kernel they were built for, we don't _have to_ include the kernel version in the installation path.

But if we don't, `multiversion(kernel)` will obviously cease to work, because the same module compiled for different kernels will use the same installation path. We will get a file conflict.

Thus _if_ we want to continue using `multiversion(kernel)`, we have to include the kernel version in the path.

Or what am I overlooking here?

hramrach:
Yes, exactly. The kernel version should be included in the path somewhere to avoid the conflict but that does not concern wm2. The only requirement is that the KMP is NOT installed someplace like /lib/modules/KVER where KVER looks like something that could potentially be a valid kernel version because in that case it's considered the old style KMP.

mwilck:
I don't think that a new argument is required. wm2 should be able to work transparently with either KMP type.

I am currently working on a patch set. The current status (yet untested) is in the PED-12049 branch.

PR #124, opened 2025-10-02
weak-modules2: skip livepatch dir when checking for unresolved symbols

mwilck:
Some livepatch modules are built without symbol versions. They are always built for one specific kernel.
Avoid that they cause false positives in the unresolved symbols check.

See https://bugzilla.suse.com/show_bug.cgi?id=1250655


hramrach:
It was re-tested and found that this is sufficient. The oldest affected release that needs this workaround is SLE15 SP6 AFAIK.

PR #128, opened 2026-04-19
Add sysconfig opt-in for insecure RNDIS USB tethering

Goainwo:
This PR provides a sysconfig variable `ALLOW_INSECURE_RNDIS` to allow users to opt-in to Android USB tethering.

Currently, `rndis_host` is blacklisted due to bsc#1205767. This breaks USB tethering for many Android devices out of the box with no user-visible error or fallback.

This patch adds:
1. A sysconfig template to expose the setting in YaST.
2. A udev rule that matches RNDIS interfaces (`bInterfaceClass=="e0"`, etc.).
3. A helper script that checks the sysconfig variable. If enabled, the udev rule explicitly `modprobe`s the module, bypassing the alias blacklist. If disabled, it logs a clear warning via `logger` to inform the user how to enable it.

This preserves the "secure by default" policy while fixing the broken UX.


mwilck:
Why can't you simply do `ln -s /dev/null /etc/modprobe.d/50-blacklist-rndis.conf` ?

Goainwo:
You're right — `ln -s /dev/null /etc/modprobe.d/50-blacklist-rndis.conf` works, and it's the standard masking pattern. For a sysadmin who already knows what's going on, it's the simplest fix.

The motivation behind this PR comes from the desktop user experience described in [bsc#1262299](https://bugzilla.opensuse.org/show_bug.cgi?id=1262299). The problem I was trying to address has two sides:

**1. Silent failure with zero feedback.** When a regular user plugs in their Android phone and enables USB tethering, nothing happens — no error, no notification, no hint in any GUI. The user has no way to discover that `rndis_host` is being blacklisted without manually inspecting modprobe configuration. Other major distributions (Fedora, Arch, Ubuntu, Debian) don't blacklist `rndis_host`, so users switching to openSUSE expect it to just work. This is especially frustrating when USB tethering is the only fallback (e.g. while troubleshooting a broken WiFi driver — which was my exact case).

**2. No easy path to enable it.** The symlink approach requires root shell access and knowledge of the modprobe override mechanism. For a home/desktop user who just wants to share their phone's internet, this is not discoverable. I was trying to provide something configurable through YaST, with clear logging (`journalctl`) when tethering is attempted while disabled — so the user at least gets a hint about *what* to do.

There's also a subtle issue: the symlink masks the entire `50-blacklist-rndis.conf`, which unblocks all three modules — including `rndis_wlan`, which is the one with the actual CVE (CVE-2023-23559). Ideally, a user who only needs USB tethering should only unblock `rndis_host`.

I also considered whether the existing `unblacklist` / `install` pattern (used for filesystem modules) could work here, but it won't — `unblacklist` requires a TTY (line 13: `tty -s <&0`) and exits silently without one. Since RNDIS loading is triggered by udev on device hotplug, there is no TTY, so the interactive prompt would never appear.

That said, I fully recognize this PR adds more infrastructure than may be appropriate, and I'm very open to simpler alternatives. Some possibilities:

- A minimal udev rule that just logs a message to `journalctl` when an RNDIS device is detected but the module is blacklisted (no sysconfig, no helper — just discoverability)
- Documentation in README.md with the symlink workaround
- Any other approach you'd prefer

I'd rather land something useful than push a specific implementation. What direction would you be comfortable with?

Goainwo:
To add to the above — I don't want to come across as if I'm attached to any particular implementation. The sysconfig/udev approach in this PR is just one way to do it, and maybe not the best one.

But I do think there's a real gap here that's worth closing: a desktop user who hits this has literally no way to figure out what's wrong without reading modprobe configs. And even once they know the cause, there's no GUI path to fix it — you need a root shell. On a distribution that ships YaST and prides itself on configurability, that feels like something we can do better.

Even if we strip this PR down to nothing else, I'd argue for at least two things:

1. Some kind of log message when an RNDIS device is detected and the driver doesn't bind — so that `journalctl` gives the user (or a forum helper) a clue. Right now it's just silence.

2. A way to toggle this from YaST or a similar tool, rather than requiring manual symlink creation. The sysconfig variable was my attempt at that, but I'm sure there are cleaner ways within the existing framework.

I'm happy to rework this into whatever shape fits the project. If you'd prefer a minimal version, or a completely different approach, just let me know — I'll do the work.

mwilck:
> 1. Silent failure with zero feedback. 

I agree that this is a problem, and we are aware of it. But it's difficult to solve.

Please look at it from a different perspective for a moment: The alternative solution discussed in the SUSE kernel team was to remove these modules completely. The enterprise customers that fund our payroll don't have much interest in these modules, so from that perspective this wouldn't hurt. SUSE knows what it owes to the community, and therefore we also care about modules that are of little interest to enterprise users, so the decision was made to keep them.

If it had been decided to remove the modules, "fixing" the tethering problem would be far more cumbersome; it would require downloading a KMP, or even compiling the module on the target system (with all the associated problems related to module signing, secure boot, MOK key handling, etc). So in a way, the blacklist approach is actually an (admittedly imperfect) attempt to be user-frienly.

There's an ongoing discussion whether we aren't actually *obliged* to remove modules which we consider potentially insecure. It is a legally dangerous situation for a company to ship kernel modules which contain known vulnerabilities, or which are not considered well-maintained in general. Just imagine for a second that you were the head of a small enterprise, and that, by auto-loading the rndis module, you'd infected your company infrastrucure with ransomware. You might rightfully consider sueing the distribution vendor for the damage.

Ubuntu and Fedora may have made different choices in this respect. These are legal / corporate decisions in which engineers have only limited say.

> I also considered whether the existing unblacklist / install pattern (used for filesystem modules)

Ok, so you've seen that. It's currently the most "user-friendly" way to configure unblacklisting that we've come up with on openSUSE. Extending this to all blacklisted modules is certainly an option, and it's low-hanging fruit, too.

> Since RNDIS loading is triggered by udev on device hotplug, there is no TTY

You are misunderstanding the mechanism. These blacklist entries exist for the very reason that automatic loading of the blacklisted modules must be avoided. In the case of file system modules, this would happen if a user plugs in e.g. a USB stick with a file system of the respective type. If that file system was maliciously crafted by an attacker, module autoloading would make it possible that malware enter the system without any explicit action on the user's side (except for plugging in the stick).

The filesystem blacklist approach *intentionally* requires the user to run e.g. `modprobe ntfs` explictly, in a terminal. Then the "unblacklist" tool would ask him whether to load the module once, or unblacklist it for the future (which would enable auto-loading).

> There's also a subtle issue: the symlink masks the entire 50-blacklist-rndis.conf, which unblocks all three modules — including rndis_wlan

The symlink is not the only possibe workaround. You can also copy the file to `/etc/udev/rules.d` and comment out just the blacklist directives that you want to avoid. We can also split the `50-blacklist-rndis.conf` file into multiple files to make the modules unblacklistable individually. I have to admit I didn't invest much thinking into that. Looking forward to your suggestions.

> No easy path to enable it.

Again, we are aware of that. For the files system blacklisting, it's [documented](https://en.opensuse.org/SDB:FilesystemBlacklisting), but not in general. And of course this documentation needs to be discovered by users, which implies that they have to figure out the right search terms.

A solution would need to cover both desktop and headless systems. The case at hand is of course most relevant for desktop systems (but someone might want to use headless RasPi or the like as tethering host).

For headless systems, I guess a journal entry plus a terminal application like `unblacklist` is about the best we can get.

For the desktop, we'd need a desktop notification telling the user that autoloading failed because of blacklisting, and giving him a hint how find further information and to eventually fix it. This mechanism would need to work on all the different desktop environments that openSUSE supports. I'm not a desktop programmer, I have no clue if such a mechanism for the situation at hand even exists. I am only aware of notifications from programs that run in the desktop user's session, not from system programs.

> A minimal udev rule that just logs a message to journalctl when an RNDIS device is detected

That sounds feasible. It would probably be even better to log this situation directly from `modprobe`. Inexperienced users wouln't notice, though, I suppose.

> Documentation in README.md with the symlink workaround

There's already quite a bit of information in the `README.md` shipped with `suse-module-tools`. If you have suggestions to improve it, patches are generally welcome. But the README is not for specific modules like `rndis_host`; after all, the method for unblacklisting is the same for all modules. There is also `/usr/lib/modprobe.d/README`. Everything documentation-related suffers from the discoverability problem, especially with inexperienced users in mind.

> On a distribution that ships YaST and prides itself on configurability, 

You may have heard that YaST is not actively maintained any more. Like it or not, it's a clear decision. Currently YaST is still used as the installer for openSUSE Tumbleweed, but that's going to change, too, sooner or later. Sorry to say it, but YaST is not the answer in 2026 any more.

> Some kind of log message when an RNDIS device is detected and the driver doesn't bind

I'd prefer something general which works for any blacklisted module.

hramrach:
> For the desktop, we'd need a desktop notification telling the user that autoloading failed because of blacklisting, and giving him a hint how find further information and to eventually fix it. This mechanism would need to work on all the different desktop environments that openSUSE supports. I'm not a desktop programmer, I have no clue if such a mechanism for the situation at hand even exists.

Did you try `notify-send`? Although there is the problem it needs to run under the user that is supposed to get the notification. I would be surprised if equivalent of `wall` for desktop did not exist.

Goainwo:
Thank you for taking the time to explain this so thoroughly — it genuinely helped me understand the reasoning behind these decisions.

A few corrections on my side:

**I was wrong about `unblacklist`.** I framed it as "won't work because there's no TTY in udev context," but that's exactly the point — the blacklist is there to *prevent* automatic loading. The workflow is: user plugs in device → nothing happens → user runs `modprobe rndis_host` manually → `unblacklist` prompts → user decides. I misunderstood the design intent. The real unsolved problem is discoverability (how does the user know to run `modprobe rndis_host` in the first place), and you've acknowledged that too.

**I didn't know YaST was being phased out.** Building around sysconfig/YaST integration was misguided — I should have researched the current direction (Agama/Cockpit) before proposing that.

**The solution was too RNDIS-specific.** You're right that a general mechanism for all blacklisted modules would be more valuable.

I also didn't appreciate the full context of the legal/corporate dimension. Knowing that the alternative was complete module removal puts the blacklist in a very different light — it's already the more user-friendly choice.

---

Based on your feedback, here's what I think I can concretely contribute with a new, minimal PR:

1. **Split `50-blacklist-rndis.conf` into per-module files** (similar to how `60-blacklist_fs-*.conf` works), so that each module (`rndis_host`, `rndis_wlan`, `usb_f_rndis`) can be individually masked or unblacklisted without affecting the others.

2. **Add the `install ... unblacklist ...` pattern for `rndis_host`**, same as the filesystem modules already use. You called this "low-hanging fruit" — I'd like to pick it.

This keeps the security default intact, gives users who run `modprobe rndis_host` in a terminal the same interactive experience that filesystem modules already have, and allows granular control without masking CVE-affected modules.

If this direction makes sense to you, I'll close this PR and open a clean one with just those changes. If you'd rather handle it differently, that's fine too — I'll follow your lead.

---

@hramrach Regarding `notify-send` — you're right that it could work, but the session-discovery problem is non-trivial (need to find the active graphical session via `loginctl`, get `DBUS_SESSION_BUS_ADDRESS`, run as the right user). It's doable but feels like a separate effort. I'd suggest tackling the `unblacklist` integration first since it's straightforward and already has precedent, and exploring desktop notifications as a follow-up — possibly as a general mechanism for all blacklisted modules, not just RNDIS.

Goainwo:
One more thing I've been thinking about, specifically regarding @hramrach's `notify-send` suggestion and @mwilck's comment that a general mechanism would be preferable.

The more I think about it, the more I believe the notification piece is where the real UX win is. The `unblacklist` integration helps users who already know to run `modprobe` — but the majority of desktop users will never get that far. They'll just think "USB tethering doesn't work on Linux" and move on.

I did some digging into how this could work as a general mechanism (not RNDIS-specific), and I see roughly three approaches with different tradeoffs:

**Approach A: structured journal entry + systemd user monitor**

The `install` directive (or `unblacklist` script) logs a structured journal message with a custom field like `SUSE_BLACKLISTED_MODULE=rndis_host`. A small systemd user service (started with the graphical session) monitors the journal for entries with that field and calls `notify-send`. This is clean — the system side doesn't need to know anything about desktop sessions, it just logs. The user side just watches the journal. General by design, works for any blacklisted module.

**Approach B: D-Bus system bus signal**

Similar to how UDisks2 emits signals when a USB drive is inserted (and file managers pick them up to show "mount this?" prompts), a small system-level helper could emit a D-Bus signal when a blacklisted module load is attempted. A desktop-side listener could then pop up a notification. More infrastructure to maintain, but the UDisks2 pattern is well-established in the Linux desktop ecosystem.

**Approach C: direct `notify-send` from system context**

Iterate `loginctl list-sessions`, find active graphical sessions, extract `DBUS_SESSION_BUS_ADDRESS` from `/run/user/$UID/bus`, run `notify-send` as that user. Technically works, but fragile and not a great pattern for a distribution-level tool.

My feeling is that Approach A is the most realistic — it has minimal moving parts and doesn't require the system side to care about desktop sessions at all. The journal is already there, structured fields are a standard mechanism, and a small user-level monitor service is easy to maintain.

But all of this is ultimately for you to decide — I'm not in a position to judge what fits the project's scope and maintenance budget. I'm happy to prototype any of these if there's interest, or to just do the minimal `unblacklist` + file-split PR and leave notifications for a future discussion.

I'll wait for your feedback before taking further action. I can close this PR whenever you say the word — I just want to make sure we've explored the options well enough to pick a good direction.

hramrach:
https://wiki.archlinux.org/title/Desktop_notifications#Send_notifications_to_all_graphical_users

Goainwo:
@hramrach Thanks — the `systemd-run --machine=user@.host --user` pattern there is clean and doesn't need any extra daemons. Good reference, I'll keep it in mind.

mwilck:
OK, so a generic notification mechanism exists. That's good.

However, let's not overdo it here. Let's begin with small steps, re-organize the rndis modules and perhaps some other blacklisted modules, too, improve the README, and make sure we get a message in the journal (as I said, I'd prefer logging this directly from `modprobe`, but I'd need to double-check the code for that; perhaps @hramrach knows if that'd be possible). Also, extend the scope of the `unblacklist` tool.

For desktop notifications, we need something generic. Like a daemon that receives system messages and forwards them to the UI sessions of logged-in users, possibly filtered by seats (e.g. in the case at hand, it might be possible to figure out that the issue was caused by plugging in a USB device that belongs to a given seat - we wouldn't want to bother users on other seats with such a message).

I see many other possible applications for such a project. Fatal system log messages such as RAID inconsistencies or imminent disk failures could be logged there, for example.

That would be a project that goes beyond module loading, and actually beyond openSUSE too. Ideally it would be approached at the systemd/logind level. (Filtering the messages in order not to overwhelm users might be the most difficult part of it, eventually).

I don't want to initimidate you. A well-engineered "small" solution just for our purpose here might well be the starting point for a more generic future project.

Goainwo:
Understood — small steps first. I'll prepare a clean PR with the RNDIS file split and `unblacklist` integration, following the existing `60-blacklist_fs-*` pattern as closely as possible.

Keeping this PR open for now since you've asked @hramrach about the `modprobe` logging possibility — I'll close it once the new one is up.

hramrach:
You can execute anything in a modprobe install command, including a tool that produces a syslog message.

Goainwo:
The clean PR is up: #129

It covers three things discussed here:
1. Split `50-blacklist-rndis.conf` into per-module files (`rndis_host`, `rndis_wlan`, `usb_f_rndis`)
2. Generalize `unblacklist` to work with any blacklisted module (not just `60-blacklist_fs-*`)
3. README update documenting non-filesystem blacklisted modules

Journal logging and desktop notifications are left for follow-up work.

Closing this PR. Thanks @mwilck and @hramrach for the guidance.

mwilck:
> You can execute anything in a modprobe install command

We already do this for file system modules, but it doesn't solve our problem.

```
> cat /usr/lib/modprobe.d/60-blacklist_fs-kafs.conf
# DO NOT EDIT THIS FILE!
#
# The kafs file system is blacklisted by default because it isn't actively
# supported by SUSE.
blacklist kafs
# The filesystem can be un-blacklisted by running "modprobe kafs".
# See README.md in the suse-module-tools package for details.
install kafs /usr/lib/module-init-tools/unblacklist kafs; /sbin/modprobe --ignore-install kafs
```

We need to log a message in the `blacklist` case, which only matters if modprobe is run for an alias, here `fs-afs`. In this case `modprobe` just exits silently, even if run verbosely.

Perhaps we could replace the blacklist statement with something like this:
```
install fs-afs /usr/bin/echo "fs-afs is aliased to kafs, which is blacklisted"
```
But that would be very clumsy for modules with lots of aliases.

Goainwo:
@mwilck That's a very important distinction — I hadn't considered the alias case. To make sure I understand correctly:

- When udev triggers `modprobe` via a module alias (e.g. `usb:v...` for RNDIS), the `blacklist` directive causes `modprobe` to exit silently — the `install` command is never reached
- The `install`/`unblacklist` mechanism only helps when the user explicitly runs `modprobe rndis_host` by name

So PR #129 addresses the "user runs modprobe directly" path (interactive un-blacklisting), but the silent-failure-on-hotplug problem would need either a change in `modprobe` itself or a separate udev-level mechanism to detect and log the situation.

Happy to help with a follow-up for the logging side once the approach is figured out.

mwilck:
Replacing `blacklist` commands by `install` commands as hinted above is not realistic, I think. Consider the list of modaliases of `rndis_host`:

```
alias:          usb:v*p*d*dc*dsc*dp*icEFisc04ip01in*
alias:          usb:v*p*d*dc*dsc*dp*icE0isc01ip03in*
alias:          usb:v*p*d*dc*dsc*dp*icEFisc01ip01in*
alias:          usb:v*p*d*dc*dsc*dp*ic02isc02ipFFin*
alias:          usb:v19D2p*d*dc*dsc*dp*ic02isc02ipFFin*
alias:          usb:v19D2p*d*dc*dsc*dp*icE0isc01ip03in*
alias:          usb:v238Bp*d*dc*dsc*dp*ic02isc02ipFFin*
alias:          usb:v1630p0042d*dc*dsc*dp*ic02isc02ipFFin*
```

We'd need to add `install` commands for all these aliases, and I don't know how to make modprobe's wildcard matching work as intended. Perhaps we'd need even more `install` commands, one for each real-world USB device matching the above wildcard expressions, which we can't even enumerate.

hramrach:
> We need to log a message in the `blacklist` case, which only matters if modprobe is run for an alias, here `fs-afs`. In this case `modprobe` just exits silently, even if run verbosely.

Sounds like a feature for kmod then.

PR #129, opened 2026-04-21
blacklist-rndis: split into per-module files and generalize unblacklist

ghost:
## Summary

Split the monolithic `50-blacklist-rndis.conf` so that each RNDIS module can be
un-blacklisted independently, and generalize the `unblacklist` tool to work
with any blacklisted module (not just filesystems).

This addresses the discoverability issue described in
[bsc#1262299](https://bugzilla.opensuse.org/show_bug.cgi?id=1262299): users who
need USB tethering with Android devices can now run `modprobe rndis_host` and
receive the same interactive un-blacklisting prompt that filesystem modules
already provide — without affecting `usb_f_rndis`.

Follows the direction discussed in #128.

### 1. Restructure RNDIS blacklist files

- **`50-blacklist-rndis.conf`** — now contains only `rndis_host` (preserving
  the original filename for backward compatibility with existing `/etc/`
  symlinks), with `install ... unblacklist ...` integration
- **`60-blacklist-usb_f_rndis.conf`** *(new)* — `usb_f_rndis` with
  `install ... unblacklist ...` integration
- **`rndis_wlan`** — removed entirely (driver was removed from the kernel
  in 6.8; blacklisting a non-existent module is misleading)

Keeping `rndis_host` in the original `50-blacklist-rndis.conf` file ensures
that users who previously un-blacklisted RNDIS via
`ln -s /dev/null /etc/modprobe.d/50-blacklist-rndis.conf` will find their
override still honoured after a package upgrade.

### 2. Generalize `unblacklist` to work with any blacklisted module

The script now supports two calling conventions:

1. `unblacklist MODULE` — locates the vendor file by pattern
   (`60-blacklist_fs-MODULE.conf` or `60-blacklist-MODULE.conf`).
   Fully backward compatible with existing filesystem blacklists.

2. `unblacklist MODULE CONFIG-BASENAME` — used when the vendor filename
   does not encode the module name (e.g. `50-blacklist-rndis.conf` for
   `rndis_host`). The basename is passed explicitly by the `install`
   directive and validated against the actual vendor file.

Also updated:
- Informational message from `"loading $MODULE file system module"` to
  `"loading module $MODULE"` (removes filesystem-specific wording)
- Regex character class from `[ \t]` to `[[:space:]]` (POSIX-portable)

### 3. Update README

Added a section documenting that non-filesystem modules can also be
blacklisted and un-blacklisted using the same mechanism.

## Not included (future work)

- **Journal logging** when a blacklisted module load is attempted
  (discussed in #128)
- **Desktop notifications** for graphical sessions (discussed in #128)

## References

- bsc#1205767, jsc#PED-5731 (RNDIS security policy)
- bsc#1262299 (USB tethering UX issue)
- Discussion: #128


modprobe.conf/common/50-blacklist-rndis_host.conf
```diff
@@ -0,0 +1,8 @@
+# DO NOT EDIT THIS FILE!
+#
+# The rndis_host module is blacklisted by default because RNDIS is
+# considered insecure (bsc#1205767, jsc#PED-5731).
+blacklist rndis_host
+# The module can be un-blacklisted by running "modprobe rndis_host".
+# See README.md in the suse-module-tools package for details.
+install rndis_host /usr/lib/module-init-tools/unblacklist rndis_host; /sbin/modprobe --ignore-install rndis_host
```
mwilck: Why use `unblacklist` here but not for the other modules?

unblacklist
```diff
@@ -25,14 +25,30 @@ if [ $# -ne 1 ]; then
...
 fi
 
-CONF=/etc/modprobe.d/60-blacklist_fs-"$MODULE".conf
+# Find the vendor configuration file that blacklists this module
+VENDOR_CONF=
+for _dir in /usr/lib/modprobe.d /lib/modprobe.d; do
+    [ -d "$_dir" ] || continue
+    for _f in "$_dir"/*.conf; do
+	[ -f "$_f" ] || continue
+	if grep -E -q "^[[:space:]]*blacklist[[:space:]]+${MODULE}[[:space:]]*$" "$_f"; then
```
mwilck: I'd rather simply extend the naming convention and check for `/usr/lib/modprobe.d/60-blacklist-"$MODULE".conf`.

mwilck: While we're at it, we don't need to check /lib/modprobe.d any more on TW.
(just a remark, you don't need to do this here).

unblacklist
```diff
@@ -55,7 +71,7 @@ elif [ -e "$CONF" ]; then
     exit 1
 fi
 
-echo "$ME: loading $MODULE file system module" >&2
+echo "$ME: *** NOTE: $MODULE will be loaded even if you answer \"n\" below. ***" >&2
```
mwilck: The simple message is sufficient, IMO.

mwilck (changes requested):
Looks good for the most part.

modprobe.conf/common/50-blacklist-rndis_host.conf
ghost: You're right — I initially kept `rndis_wlan` and `usb_f_rndis` as plain blacklists because of CVE-2023-23559 for `rndis_wlan`, but on reflection, the `unblacklist` mechanism already requires explicit user action and confirmation. It should be consistent across all three. Will fix.

unblacklist
ghost: Agreed — the naming convention approach is cleaner and more predictable. I'll rename the files to `60-blacklist-MODULE.conf` and replace the grep loop with a simple check for both `60-blacklist_fs-` and `60-blacklist-` patterns.

unblacklist
ghost: Will simplify.

Goainwo:
Pushed the fixes addressing all three review comments:

1. **All three modules now use `install ... unblacklist`** — consistent with how filesystem blacklists work. The user is always asked before un-blacklisting, so this doesn't weaken security.

2. **Naming convention extended instead of grep** — files renamed from `50-blacklist-*` to `60-blacklist-MODULE.conf`, and the `unblacklist` script now simply checks for both `60-blacklist_fs-MODULE.conf` and `60-blacklist-MODULE.conf` patterns. Much simpler.

3. **Simple message** — changed to `"loading module $MODULE"`, no verbose NOTE format.

The net diff from master should now be clean:
- `50-blacklist-rndis.conf` → deleted (original monolithic file)
- `60-blacklist-rndis_host.conf` → added (with unblacklist)
- `60-blacklist-rndis_wlan.conf` → added (with unblacklist)
- `60-blacklist-usb_f_rndis.conf` → added (with unblacklist)
- `unblacklist` → extended with second naming pattern
- `README.md` → new section for non-filesystem blacklisted modules

mwilck:
No offense, but @Goainwo, are you, by any chance, an AI bot?
Not that it means much, but I'd appreciate being honest about this.

Goainwo:
No offense taken — fair question.

I'm a real person (Oleksandr, based in Ukraine). I do use AI tools to help with research, drafting, and iterating faster — which probably explains the response speed and the somewhat polished formatting you may have noticed.

But every change, every reply, and every design decision goes through me. I read your feedback, think about it, verify the code, and decide what to do. The mistakes in the earlier iterations (the overengineered grep approach, the inconsistent unblacklist coverage) were mine — the AI doesn't have the domain context to catch those, and I should have thought more carefully before pushing.

I appreciate the honest and patient review. Happy to continue working on this the right way.

mwilck:
Thanks. I can't quite cope with the speed of your responses though. I'll do another review iteration tomorrow.

modprobe.conf/common/60-blacklist-rndis_host.conf
```diff
@@ -0,0 +1,8 @@
+# DO NOT EDIT THIS FILE!
```
mwilck, line 1: Actually, there is a problem with this splitting of `50-blacklist-rndis.conf` which I haven't realized so far: A user who has previously un-blacklisted the rndis drivers would now find them blacklisted again. Such users would be justifiably grumpy.

I am uncertain what to do about this.

While this new file has lower "priority" than `50-blacklist-rndis.conf`, it will still be effective for users that have applied the `ln -s /dev/null` technique, or for users who have simply commented out the `blacklist` directives for the rndis drivers.

The `rndis_wlan` driver has been removed in kernel 6.8 and is thus not available any more in TW and Leap 16. It makes no difference if it's blacklisted or not.

The remaining rndis drivers are used for complementary tasks: `rndis_host` for using a mobile phone as network connection from the Linux host, and `usb_f_rndis` as gadget driver, using the linux host as network device for a mobile device or for a Windows host (correct me if I'm wrong).

I believe that it's `rndis_host` that matters most. Thus, perhaps we should rename `60-blacklist-rndis_host.conf` to `50-blacklist-rndis.conf`. That would  require changes to the conf-file detection in the `unblacklist` tool, too.

hramrach: Also there is a message which is displayed when a new version of a package is installed which could be used to notify the user.

I carrying the exception long-term is not great.

ghost: Good catch — I didn't consider the upgrade path for users who have already un-blacklisted rndis via `ln -s /dev/null /etc/modprobe.d/50-blacklist-rndis.conf`. Their symlink would be ineffective against the renamed files.

A few observations:

**1. `rndis_wlan` can be dropped entirely.** Confirmed it was removed from the kernel in 6.8 — there's no module to blacklist on TW or Leap 16.

**2. Keeping both modules in one file creates a coupling problem with `unblacklist`.** If `50-blacklist-rndis.conf` contains both `rndis_host` and `usb_f_rndis`, then:
- Answering "yes" for `rndis_host` creates a `/dev/null` symlink that masks the entire file → `usb_f_rndis` is un-blacklisted too (side effect)
- Answering "never" for `rndis_host` creates an `/etc/` file with only `blacklist rndis_host` → the `/etc/` version replaces the vendor file → `usb_f_rndis` loses its blacklist entry (bug)

So per-module files are needed for `unblacklist` to work independently, which is why the filesystem blacklists use one file per module.

**3. The rename creates a name mismatch for `unblacklist`.** If we keep the file as `50-blacklist-rndis.conf` but the module is `rndis_host`, the pattern-based detection (`60-blacklist-MODULE.conf`) won't find it. One clean way to handle this: let the `install` directive pass the config basename as an optional second argument:

```
install rndis_host /usr/lib/module-init-tools/unblacklist rndis_host 50-blacklist-rndis; /sbin/modprobe --ignore-install rndis_host
```

The script would check: if `$2` is provided, use it as the config basename; otherwise fall through to the existing pattern checks. This is backward compatible — existing filesystem `install` directives don't pass a second argument, so they continue to work via patterns.

**Proposed structure:**
- `50-blacklist-rndis.conf` → only `rndis_host` + install/unblacklist (preserves backward compat for the most important module)
- `60-blacklist-usb_f_rndis.conf` → `usb_f_rndis` + install/unblacklist (new file, no backward compat concern)
- No file for `rndis_wlan` (removed from kernel)

For the edge case where a user previously un-blacklisted everything via the symlink: `rndis_host` stays un-blacklisted (symlink still works ✅), but `usb_f_rndis` would become blacklisted again via the new file. Since `rndis_host` is the primary use case, this seems like an acceptable trade-off — but happy to hear your thoughts.

Regarding @hramrach's point about RPM update messages — that could additionally be used to notify users about the `usb_f_rndis` change, though it's likely a very small number of users.

Let me know which direction you'd prefer and I'll implement it.

Goainwo:
Thanks for pointing out the upgrade path issue for users with existing `50-blacklist-rndis.conf` symlinks. That was indeed a blind spot in the initial split.

I've pushed an update that addresses this and incorporates the feedback from this thread:

1. **Renamed `60-blacklist-rndis_host.conf` back to `50-blacklist-rndis.conf`**
   This preserves the upgrade path. Users who previously un-blacklisted RNDIS by symlinking `/etc/modprobe.d/50-blacklist-rndis.conf` to `/dev/null` will find their override still honored after a package upgrade. This file now *only* contains `rndis_host`, so the per-module unblacklist operations remain completely independent and do not cause coupling bugs.

2. **Dropped `60-blacklist-rndis_wlan.conf` entirely**
   As you noted, `rndis_wlan` was removed from the kernel in 6.8. Since it's no longer present in Tumbleweed or Leap 16, blacklisting it is harmless but misleading, so I've removed it from the package.

3. **Extended `unblacklist` to accept an optional config-basename argument**
   Since the `rndis_host` vendor file is now named `50-blacklist-rndis.conf`, the strict `60-blacklist-$MODULE.conf` naming convention won't find it. To solve this cleanly, the `install` directive in `50-blacklist-rndis.conf` now explicitly passes the basename to the script:
   `install rndis_host /usr/lib/module-init-tools/unblacklist rndis_host 50-blacklist-rndis; ...`
   
   The `unblacklist` script checks if this second argument is provided and validates that the referenced vendor file actually exists to guard against typos. All existing callers (filesystem blacklists, and the new `60-blacklist-usb_f_rndis.conf`) remain fully backward-compatible as they pass no second argument and fall through to the existing pattern checks.

This approach keeps `rndis_host` completely backward-compatible for existing users, correctly isolates `usb_f_rndis`, removes the dead `rndis_wlan` module, and keeps the `unblacklist` script robust and generalized.

Let me know if this looks good to you!

Goainwo:
Pushed one more fix: the previous README.md changes accidentally included unrelated formatting edits (removed headings, blank lines, changed bullet point indentation). This commit restores the original formatting and keeps only the new "Other blacklisted modules" section.

I've also updated the PR description to reflect the current state of the changes.

The commit history is a bit noisy from the review iterations — happy to squash into clean logical commits if/when you're satisfied with the current state.

Goainwo:
Squashed the review iteration history into two clean commits:

1. **`unblacklist: generalize to support non-filesystem-specific modules`** — extends the script with the optional `config-basename` argument and the `60-blacklist-MODULE.conf` pattern check.

2. **`modprobe.conf: split RNDIS blacklist, add interactive unblacklist support`** — restructures the RNDIS blacklist files, drops `rndis_wlan`, and documents the mechanism in README.

No functional changes from the previous state — just a clean history for easier review.

mwilck (approved):
LGTM

PR #130, opened 2026-05-07
Support XBOOTLDR

aplanas:


weak-modules2
```diff
@@ -480,7 +480,9 @@ get_current_initrd() {
     if [ -n "$SDBOOTUTIL" ]; then
 	if [ -z "$TRANSACTIONAL_UPDATE" ] && [ -s '/etc/kernel/entry-token' ]; then
 	    read -r _ initrd <<<"$(sdbootutil --entry-keys=initrd show-entry $krel)"
-	    initrd="/boot/efi$initrd"
+	    # The initrd can be in the ESP (/boot/efi, /boot) or in
+	    # the XBOOTLDR (/boot), depending on the installation
+	    initrd="$(bootctl --print-boot-path)$initrd"
```
mwilck, line 485: What if bootctl is not installed? Can you add a fallback for that case?

mwilck (changes requested):

weak-modules2
aplanas: `bootctl` is part of `udev` package, and is a hard requirement of `sdbootutil`, so if `bootctl` is missing then there is a big issue without fallback

mwilck: Ok. I missed the fact that this is in an `if [ -n "$SDBOOTUTIL" ]` block.

mwilck (approved):

